PR #7, opened 2015-12-09
Adding support for Telegraf line protocol

mmoulton:
This PR adds support for InfluxDB's Telegraf [line protocol](https://github.com/influxdb/telegraf/tree/master/plugins/statsd). The primary difference is in how tags are sent, as well as InfluxDB does not have support for events.

I took the liberty of bumping the version (assuming semver) and updating the changelog. Let me know if you would rather I pull these out for you to update.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -16,7 +16,7 @@ var dgram = require('dgram'),
  *   @bufferFlushInterval {Number} the time out value to flush out buffer if not
  * @constructor
  */
-var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock, globalTags, maxBufferSize, bufferFlushInterval) {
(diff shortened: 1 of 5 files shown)


lib/statsd.js
```diff
@@ -266,7 +272,7 @@ Client.prototype.sendStat = function (stat, value, type, sampleRate, tags, callb
       message += '|@' + sampleRate;
     } else {
       //don't want to send if we don't meet the sample ratio
-      return;
+      return callback();
```
bdeitte: Thanks for the additional fixup here

bdeitte:
This looks great, thanks.  I think some additional things need to be worked around in the README to make clear what's allowed in different cases.  I see some other changes I should make in there though, so I'm just going to make those in a followup checkin.  I should get to that shortly and will then publish the new version.

PR #11, opened 2015-12-14
Telegraf Bug Fix

mmoulton:
Fixed bug in search and replace where only first tag was formatted correctly for telegraf line protocol.


bdeitte:
LGTM, will get a patch version out with this

PR #14, opened 2016-02-03
Bug Fix: Should not send messages larger then maxBufferSize

Pchelolo:
Currently the queue is flushed when the buffer size is larger then `maxBufferSize`. However, this results in emitting messages larger then maximum size. As max size should be more or less equal to the network MTU, this results in package truncation in the network. So, instead we need to flush the queue when we discover that adding one more message to the buffer will result in exceeding the max buffer size.


bdeitte:
Makes sense, thanks for the update.  I'll get a new version out with this shortly.

PR #15, opened 2016-02-24
Ignore some more things when publishing

arlolra:
https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package


bdeitte:
Thanks for the additions, makes sense.

PR #16, opened 2016-02-25
Add a create method that does dns caching before constructing

arlolra:
/cc @gwicke

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -68,6 +68,23 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock, gl
 };
 
 /**
+ * Callback with a constructed client,
+ * returning an error for a dns cache failure.
+ */
+Client.create = function(options, cb) {
+  options = options || {};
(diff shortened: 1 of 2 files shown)


bdeitte:
Thanks for the change here.  I'm not excited about having a different creation method, given there's essentially two ways to do already (with the unusual dual way that options can be passed in).  But I don't have any better answer myself to add this functionality without breaking backwards compatibility, which I don't want to do to allow node-statsd users to keep migrating.  So looks like we're stuck with this!  

I'm going to add some info to the README about this after merging sometime today, and then I'll get a new version published with both your changes in it.

gwicke:
Another option would be to drop metrics on the floor if resolution has failed, and report an error in the (optional) callback. Often, not exploding is all that's needed.

bdeitte:
Great point @gwicke.  That does make sense to me.  I made those changes instead: https://github.com/brightcove/hot-shots/pull/17

I'll merge this in sometime tomorrow and version on top of that change if there's no objections before then.

PR #18, opened 2016-03-07
Test the latest along with everything else

bdeitte:


bdeitte:
As I saw in https://github.com/sivy/node-statsd/pull/63

PR #20, opened 2016-04-26
Child client

ash2k:
A child client can be used e.g. in a middleware - it can be passed to the request handlers, enriched with additional tags. WDYT?

Inspired by https://github.com/msiebuhr/node-statsd-client/blob/master/lib/statsd-client.js#L40

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -49,13 +49,15 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock, gl
   this.bufferFlushInterval = options.bufferFlushInterval || 1000;
   this.bufferHolder = options.isChild ? options.bufferHolder : { buffer: "" };
 
+  // We only want a single flush event per parent and all its child clients
   if(!options.isChild && this.maxBufferSize > 0) {
     this.intervalHandle = setInterval(this.onBufferFlushInterval.bind(this), this.bufferFlushInterval);
   }
 
   if (options.isChild) {
-    this.cacheDns = options.cacheDns;
-    this.dnsError = options.dnsError;
+    if (options.dnsError) {
+      this.dnsError = options.dnsError;
+    }
   } else if (options.cacheDns === true){
     dns.lookup(options.host, function(err, address, family){
       if(err === null){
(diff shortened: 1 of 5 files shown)


lib/statsd.js
```diff
@@ -40,24 +41,27 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock, gl
...
+  this.bufferHolder = options.isChild ? options.bufferHolder : { buffer: "" };
 
-  if(this.maxBufferSize > 0) {
+  if(!options.isChild && this.maxBufferSize > 0) {
     this.intervalHandle = setInterval(this.onBufferFlushInterval.bind(this), this.bufferFlushInterval);
   }
 
-  if(options.cacheDns === true){
+  if (options.isChild) {
+    this.cacheDns = options.cacheDns;
```
bdeitte: cacheDns isn't set on this in the non-child case, so I don't think you need to set it here.  I would think though you want to grab the 'host' from the parent if it's available and set it, since that is done in the skipped 'else if' block below this.

lib/statsd.js
```diff
@@ -40,24 +41,27 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock, gl
   this.port        = options.port || 8125;
   this.prefix      = options.prefix || '';
   this.suffix      = options.suffix || '';
-  this.socket      = dgram.createSocket('udp4');
+  this.socket      = options.isChild ? options.socket : dgram.createSocket('udp4');
```
bdeitte: Can you add a comment to the ones you are overriding as to why they need to be overriden for the child case?  I think I get why for each of them, but a comment would help folks first looking at this as to why only some get this treatment.

bdeitte:
Thanks for the addition here!  Makes sense to me add this in- I haven't done this before myself with statsd but can see the need.  (I have done something similar with bunyan before.)  I have a few small comments, and could you also add something in the README.md about this?  After that I can merge and create a new release.

lib/statsd.js
ash2k: Fixing this. Also note that `childClient()` method does get `host` from parent client.

ash2k:
@bdeitte please let me know if you want me to do any additional tweaks. I've added some comments, updated docs.

bdeitte:
Looks great, thanks

bdeitte:
While it's not technically a "major release" for changing functionality, it's a big enough change for what usually changes in a minor for this project that I'll create a 3.0.0 now you'll see this in.

PR #23, opened 2016-04-29
Error handler

ash2k:
Libraries should not assume that they could write to `console`. This PR provides a way to override default behaviour and also provide a handler for errors when `callback` was not passed. Also some minor cleanups.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[**]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+trim_trailing_whitespace = true
+
+[**.{json,js,yml,yaml}]
+indent_style = space
+indent_size = 2
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -15,6 +15,7 @@ var dgram = require('dgram'),
  *   @option cacheDns    {boolean} An optional option to only lookup the hostname -> ip address once
  *   @option mock        {boolean} An optional boolean indicating this Client is a mock object, no stats are sent.
  *   @option globalTags {Array=} Optional tags that will be added to every metric
+ *   @option errorHandler {Function=} Optional function to handle errors when callback is not provided
  *   @maxBufferSize      {Number} An optional value for aggregating metrics to send, mainly for performance improvement
  *   @bufferFlushInterval {Number} the time out value to flush out buffer if not
  * @constructor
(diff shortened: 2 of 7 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
```
ash2k: Makes Idea and other editors recognise the style used in the project. Very useful.

lib/statsd.js
```diff
@@ -289,7 +302,7 @@ Client.prototype.sendStat = function (stat, value, type, sampleRate, tags, callb
       message += '|@' + sampleRate;
     } else {
       //don't want to send if we don't meet the sample ratio
-      return callback();
+      return callback ? callback() : undefined;
```
ash2k: `callback` is an optional parameter - must check if it was passed.

lib/statsd.js
```diff
@@ -299,7 +312,7 @@ Client.prototype.sendStat = function (stat, value, type, sampleRate, tags, callb
  * Send a stat or event across the wire
  * @param message {String} The constructed message without tags
  * @param tags {Array} The tags to include (along with global tags). Optional.
- * @param callback {Function=} Callback when message is done being delivered. Optional.
+ * @param callback {Function=} Callback when message is done being delivered (only if maxBufferSize == 0). Optional.
```
ash2k: It is not called in the buffered output case.

test/test_statsd.js
```diff
@@ -556,9 +556,42 @@ function doTests(StatsD) {
...
+            finished();
+          });
+        });
+      });
+    });
+
   });
 
-  describe('#gauge', function(finished){
+  describe('#gauge', function(){
```
ash2k: `describe` does not provide callback, only `it`.

test/test_statsd.js
```diff
@@ -1071,50 +1114,81 @@ function doTests(StatsD) {
...
+          assert.equal(e, err);
           finished();
-        });
+        }
       });
+      statsd.dnsError = err;
+      statsd.send('test title');
     });
 
-    it('should have error in callback after bad histogram call', function(finished){
```
ash2k: I moved histogram tests into `#histogram`.

bdeitte:
Thanks for the PR.  I should get to review this tonight.

bdeitte:
Could you add the details to README.md?  Everything else looks fine to me.

I am curious though- in what case do you not have access to the console?  I never thought of this running outside of Node.js, so I am curious on the use case here.  Or do you mean that you just want to make sure the console messages go to a proper log no matter what?

ash2k:
I'll add changes to README.

Yes, I want all messages to go to the logger (which writes them to console in JSON format).

bdeitte:
Looks good, thanks for another addition.

PR #24, opened 2016-05-03
Override matching tags

### LINKED ISSUE #22: Child client should override matching tags
If parent client already has some global tags, client's tags should override tags with same name (case-sensitive?).

ash2k:
Fixes #22

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -327,12 +327,9 @@ Client.prototype.send = function (message, tags, callback) {
     throw this.dnsError;
   }
 
-  var mergedTags = [];
+  var mergedTags = this.globalTags;
   if(tags && Array.isArray(tags)){
-    mergedTags = mergedTags.concat(tags);
-  }
-  if(this.globalTags && Array.isArray(this.globalTags)){
-    mergedTags = mergedTags.concat(this.globalTags);
+    mergedTags = overrideTags(mergedTags, tags);
   }
   if(mergedTags.length > 0){
     if (this.telegraf) {
@@ -472,7 +469,8 @@ var ChildClient = function (parent, options) {
     globalize   : false, // Only "root" client can be global
     mock        : parent.mock,
     // Append child's tags to parent's tags
-    globalTags  : Array.isArray(options.globalTags) ? parent.globalTags.concat(options.globalTags) : parent.globalTags,
+    globalTags  : Array.isArray(options.globalTags) ?
+        overrideTags(parent.globalTags, options.globalTags) : parent.globalTags,
     maxBufferSize : parent.maxBufferSize,
     bufferFlushInterval: parent.bufferFlushInterval,
     telegraf    : parent.telegraf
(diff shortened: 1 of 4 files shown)


lib/statsd.js
```diff
@@ -464,5 +462,37 @@ Client.prototype.childClient = function(options) {
   return new ChildClient(this, options);
 };
 
+// Overrides tags in parent with tags from child with the same name (case sensitive) and return the result as new
+// array. parent and child are not mutated.
```
bdeitte: Nitpick, but can you change this to be a /* */ style comment like the other function-level comments?

lib/statsd.js
```diff
@@ -464,5 +462,37 @@ Client.prototype.childClient = function(options) {
...
+function overrideTags (parent, child) {
+  var childCopy = {};
+  var toAppend = [];
+  child.forEach(function (tag) {
+    var idx = tag.indexOf(':');
```
bdeitte: You can (accidentally) add null or undefined to an array, so there should be a check here for this

lib/statsd.js
```diff
@@ -464,5 +462,37 @@ Client.prototype.childClient = function(options) {
...
+    if (idx < 1) { // Not found or first character
+      toAppend.push(tag);
+    } else {
+      childCopy[tag.substring(0, idx)] = tag.substring(idx + 1);
+    }
+  });
+  var result = parent.map(function (tag) {
+    var idx = tag.indexOf(':');
```
bdeitte: Same comment about null/undefined

lib/statsd.js
```diff
@@ -311,12 +311,9 @@ Client.prototype.send = function (message, tags, callback) {
...
   }
 
-  var mergedTags = [];
+  var mergedTags = this.globalTags;
   if(tags && Array.isArray(tags)){
-    mergedTags = mergedTags.concat(tags);
-  }
-  if(this.globalTags && Array.isArray(this.globalTags)){
-    mergedTags = mergedTags.concat(this.globalTags);
+    mergedTags = overrideTags(mergedTags, tags);
```
bdeitte: I believe this changes things not just for child clients but also for adding tags over global tags.  I'm fine with this, it makes sense to me to do this.  I will though bump it to a major version for this though as it is a change in functionality.

bdeitte:
A few small comments, and then I'll merge, thanks.

lib/statsd.js
ash2k: Yes, it'll change the behaviour but I did it intentionally. I think it makes sense.

lib/statsd.js
ash2k: Well, if you mean `tag` could be `null` or `undefined` then IMO it should blow up because it is a [programming error](https://www.joyent.com/developers/node/design/errors). That way user can fix the bug in his/her code and it is better than adding meaningless tags. WDYT?

ash2k:
Rebased onto master, updated `CHANGES.md`, fixed docs comment style.

lib/statsd.js
bdeitte: Joyent's article, in my opinion, is saying that you should not continue after having an unhandled exception.  I very much agree with that in how Node.js works.  I don't agree though that you should allow unhandled exceptions to happen if there is a way to prevent them.  This causes a lot of requests/activities to stop their work potentially and can be a confusing thing to deal with.  

I am fine with this being one of the error cases where we throw an Error (or use the error handler) but would really rather not have it throw an unexpected error.  Especially in the send() function for existing merging of tags, where this is changing existing programs that use them and could suddenly throw uncaught exceptions.

ash2k: There is always a way to "avoid" unexpected exceptions - check each and every parameter for correctness and try to do something meaningful with it. I think as a general approach to writing code it is a bad idea: 
# user of the library should use it correctly and if it is misused it should let the user know about it.
# the code becomes full of "safety checks" and it makes it hard to maintain.
# There is no guarantee all cases are handled anyway (bug in a library itself may cause an exception).

I agree that this changes the behaviour but I thought you wanted to release a new major version? :) This feature itself changes the behaviour for existing programs.

p.s. as a user of the library I'd prefer the library to fail fast and crash my program to let me know that I'm doing something stupid so that I can go and fix my crappy code, rather than masking the bug. Running program that produces no errors but incorrect results is a nightmare to debug, isn't it?

bdeitte: I think we're on close to the same idea here but aren't quite finding the common path.  :)  I agree that having no errors would not be the best solution- I just really, really don't want uncaught exceptions to be the norm, just based on my own experience with Node projects in production.  So all I ask is that the error is returned in a callback (or in the fallback error handler) instead of being an uncaught exception.  Does that work for you?

ash2k:
Hey, just letting you know that I've updated the PR. Have a good weekend!

bdeitte:
Thanks, have a good weekend too!

PR #32, opened 2016-09-30
Allow socket errors to be handled with errorHandler. Fixes #31.

### LINKED ISSUE #31: No `error` listener is added to the udp socket
Running `hot-shots` in production in a kubernetes environment results in sporadic dns lookup failures for me. As there is no `error` listener attached to the socket my hypothesis is that dns lookup errors are bubbling their way up as uncaught exceptions, resulting in a server crash.

Request: Use the existing `options.errorHandler` for:

```js
if (options.errorHandler) {
  this.socket.on('error', options.errorHandler);
}
```

ggoodman:
This PR:

- Uses the optional `errorHandler`, if present, as an `error` listener on the udp socket (**only on the parent**.
- Adds tests demonstrating scenarios that are and are not handled.

What is unclear:

- Should the `errorHandler` of child clients also be added as listeners? Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,9 +164,19 @@ As usual, callbacks will have an error as their first parameter.  You can have a
 
 If the optional callback is not given, an error is thrown in some cases and a console.log message is used in others.  An error will only be thrown when there is a missing callback if it is some potential configuration issue to be fixed.
 
-In the event that there is a socket error, `hot-shots` will allow this error to bubble up.  If you would like to catch the errors, just attach a listener to the socket property on the instance.
+In the event that there is a socket error, `hot-shots` will allow this error to bubble up unless a `errorHandler` is specified.  If you would like to catch the errors, either specify an `errorHandler` in your root client or just attach a listener to the socket property on the instance.
 
 ```javascript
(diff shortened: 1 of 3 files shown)


ggoodman:
Travis failed on `npm install` in 2nd commit. Tests appear to pass locally and in 3/4 environments.

bdeitte:
I reran the failing Travis build.  Odd blip and passing now.

Given that the socket is shared between all children and the parent, I think you have it right in there and that child clients should not listen in.

Can you add to the Errors section in the README and explain this change there?  I completely forgot that sockets were already mention in there.  (Before my time.)  I would probably just replace the last paragraph and the code snippet in there and consider this errorHandler to be the best practice now, so say something like this instead:

"In the event that there is a socket error, `hot-shots` will allow this error to bubble up.  If you do not want this to happen, use an errorHandler to listen for this error."

ggoodman:
I also totally missed that section in the README. I'll adjust accordingly.

bdeitte:
Great, merging and will tag a build soon afterwards.

PR #42, opened 2017-05-04
Support default value with tags in #increment

### LINKED ISSUE #33: Calling increment with only args results in corrupted packets
```
    dogstatsd.increment('express.response.error', [
      `url:${(req.route && req.route.path) || Sanitize.url(req.url)}`,
      `statusCode:${req.statusCode}`
    ]);
```

I had assumed that the above would default to value of 1, but it does not, resulting in error in the dogstats agent.

```
2016-10-24 10:41:09 CDT | ERROR | dd.dogstatsd | dogstatsd(dogstatsd.py:366) | Error receiving datagram
Traceback (most recent call last):
  File "/opt/datadog-agent/agent/dogstatsd.py", line 354, in start
    aggregator_submit(message)
  File "/opt/datadog-agent/agent/aggregator.py", line 608, in submit_packets
    parsed_packets = self.parse_metric_packet(packet)
  File "/opt/datadog-agent/agent/aggregator.py", line 465, in parse_metric_packet
    raise Exception('Unparseable metric packet: %s' % packet)
Exception: Unparseable metric packet: express.response.error:url:/not-found,statusCode:null|c
```

Adding a conditional like the following seems to fix the issue

```
    if (arguments.length < 3) {
      if (typeof value !== 'number') {
        tags = value;
        value = undefined;
      }
    }
```

jsocol:
Fixes #33.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -113,6 +113,13 @@ Client.prototype.timing = function (stat, time, sampleRate, tags, callback) {
  * @param callback {Function=} Callback when message is done being delivered. Optional.
  */
 Client.prototype.increment = function (stat, value, sampleRate, tags, callback) {
+  // allow use of tags without explicit value or sampleRate
+  if (arguments.length < 3) {
+    if (typeof value !== 'number') {
+      tags = value;
+      value = undefined;
+    }
+  }
   // we explicitly check for undefined and null (and don't do a "! value" check)
   // so that 0 values are allowed and sent through as-is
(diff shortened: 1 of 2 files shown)


jsocol:
I took the code directly from #33, but I'm not sure if this will correctly handle a case where value is defined but sampleRate is not, adding another test!

bdeitte:
Looks great, thanks!

jsocol:
Thank you!

PR #46, opened 2017-09-19
Typescript typings

jgwmaxwell:
Hi there,

I wonder if you'd be open to adding Typescript typings to this library? If you aren't familiar with it, this just takes the form of a definition file that is pointed to from `package.json`, it doesn't require a comprehensive reworking, or in fact any changes to the source code.

I know that maintaining a types file is a little extra work, but the API of this library seems stable enough that it might not be onerous, and I'd be happy to help keep them in sync if required.

Hope this is ok,

John

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "description": "Node.js client for StatsD, DogStatsD, and Telegraf",
   "version": "4.5.0",
   "author": "Steve Ivy",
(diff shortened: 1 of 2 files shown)


bdeitte:
I'm for it- I use TypeScript at work right now myself.  We just haven't gotten to something like this, happy to see you have!

PR #51, opened 2017-10-31
concat prefix and suffix in check function

### LINKED ISSUE #38: global prefix is not concatated for "check" function
Hi,
Thanks for this library, very useful.
calling "check" function like this:
`client.check('service.running', this.CHECKS.OK, [this.serviceName]);`
with global prefix param set to "somePrefix." 
I would expect the data sent to Datadog to be named as "somePrefix.service.running" like happends in other functions but isn't.

Jiggmin:
Fixes #38

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -201,7 +201,8 @@ Client.prototype.check = function(name, status, options, tags, callback) {
 
     throw err;
   }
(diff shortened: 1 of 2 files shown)


bdeitte:
I will merge this PR and the other one late today or tomorrow- thanks for all the contributions!

Issue #48, opened 2017-09-30
Typescript typings


bdeitte:
Add back typescript typings as noted in https://github.com/brightcove/hot-shots/pull/46 Making an issue in case @jgwmaxwell doesn't get to it, or if that does happen assigning the hacktoberfest label so he can get credit for it if he wants.  :)

karsai5:
In order to get the typescript definitions to work I had to change the module export. I've created a pull request at https://github.com/brightcove/hot-shots/pull/54

PR #52, opened 2017-11-04
Typescript types take 2

leemhenson:
> This change is not backwards compatible with Typescript projects that are already using statsd.
>
> require('hot-shots') does not resolve to StatsD
> the client (StatsD instance) is missing the socket method

I've added some missing options, but I'm unsure how to resolve the the default export issue. Whenever I try to do `export = StatsD` I get an error:

`error TS2309: An export assignment cannot be used in a module with other exported elements.`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "description": "Node.js client for StatsD, DogStatsD, and Telegraf",
   "version": "4.8.0",
   "author": "Steve Ivy",
+  "types": "./types.d.ts",
   "contributors": [
     "Russ Bradberry <rbradberry@gmail.com>",
     "Brian Deitte <bdeitte@gmail.com>",
--- a/types.d.ts
+++ b/types.d.ts
@@ -1,0 +1,74 @@
+declare module "hot-shots" {
+  export interface ClientOptions {
+    host?: string;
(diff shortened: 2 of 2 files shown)


leemhenson:
Ping https://github.com/brightcove/hot-shots/pull/46 and https://github.com/brightcove/hot-shots/pull/48

bdeitte:
Thanks!  I am concerned about it not being backwards compatible, or really understanding if it's not possible (or not the right).  @sandy-adi do you have any thoughts here?

leemhenson:
@bdeitte I think this now covers the default export too.

bdeitte:
@leemhenson Awesome, thank you!  Here's what I'm going to do here:

1) Try this out myself on a TypeScript project I work on to make sure it doesn't break it
2) Wait another day or two to let anybody else try it out
3) Tag it as a major release (with a note that it's not supposed to be semver-major, but just in case)

bdeitte:
I did try this out and run through all of our tests in a TypeScript project, and it worked just fine.  I'll still be waiting until likely Thursday morning, to see if @sandy-adi or anybody else wants to this out then will get this out there!

PR #56, opened 2018-02-02
Add childClient to types

punya-asapp:
This useful method is already defined and used in the README, but not yet reflected in the typings.


bdeitte:
Good catch, thanks!

punya-asapp:
Thanks for the quick turnaround! Do you happen to know when this will show up in a release (so I can pick it up in my project)?

bdeitte:
Already done!  5.0.1

PR #58, opened 2018-02-13
Pass key-value tags as objects

### LINKED ISSUE #55: Pass key-value tags as objects
To my understanding, the StatsD implementations with tag support expect the tags to be key-value pairs. Hot-shots expects users to do the formatting, which is not ideal. Users must also do sanitation of tags which is even less ideal.

Instead, there could be automatic formatting of key-value tags to array

```javascript
statsd.increment("metric", 1, { "key": "value" });
```

should be the same as

```javascript
statsd.increment("metric", 1, ["key:value"]);
```

The implementation could be backwards compatible. If tags is passed as an array, no formatting would be performed.

[InfluxDB](https://www.influxdata.com/blog/getting-started-with-sending-statsd-metrics-to-telegraf-influxdb/) and [Datadog Statsd](http://docs.datadoghq.com/guides/dogstatsd/) use different format for tags. This annoying, but could be addressed by a configuration option.

I've created [a wrapper](https://github.com/lautis/hot-shots-deux) for hot-shots that implements the tagging for the Datadog format. Would you be interested in having this functionality in hot-shots? I'd be happy to make a PR out of this.

lautis:
Implement passing key-value tags as objects. This reduces the chance of errors, and also allows implementing sanitation of tags.

The old format is still supported, and this is relying on the existing parsing and formatting functionality. If the array format would be removed, some of the parsing would become redundant and tags could be passed as an object until they are serialised as a string over the wire.

I'm not at all sure about the escaped characters: is the list complete and are there redundant characters? The documentation for this seems to be non-existent as even the reference implementations use arrays instead of objects.

Fixes #55.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Parameters (specified as an options hash):
 * `globalize`:   Expose this StatsD instance globally? `default: false`
 * `cacheDns`:    Cache the initial dns lookup to *host* `default: false`
 * `mock`:        Create a mock StatsD instance, sending no stats to the server? `default: false`
-* `globalTags`:  Tags that will be added to every metric `default: []`
+* `globalTags`:  Tags that will be added to every metric. Can be either an object or list of tags. `default: {}`
 * `maxBufferSize`: If larger than 0,  metrics will be buffered and only sent when the string length is greater than the size. `default: 0`
 * `bufferFlushInterval`: If buffering is in use, this is the time in ms to always flush any buffered metrics. `default: 1000`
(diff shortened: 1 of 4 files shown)


bdeitte:
This looks great to me- thanks for putting this together!  Could you add a note in the README.md as part of this PR as well, noting the two styles of tags now allowed?  

I don't know the answer to the escaped characters either- I'm fine personally with having this go in and iterating on it if someone does come up with an answer here.  I'll wait until late tomorrow in any case to see if anybody happens to see this for an answer, then I can merge this in and make a minor release of it.

lautis:
Updated readme.

PR #59, opened 2018-02-16
Add 'timer' decorator function

ericapisani:
Some of my co-workers and I have been finding it very useful to record how long a given function takes to execute, and then send that time on to our StatsD server using the `timing` function. We thought this might be a useful addition to the library.

Sample use of this function:

```
function add(a, b) {
  return a + b;
}

// Would send the time taken to execute 'add' as the 'time'
// argument to the 'timing' function
timer(add, 'add_execution_time')(2, 3);

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,8 +91,7 @@ The check method has the following API:
   // Timer: Records how long a given function takes to execute (in milliseconds),
   // and then sends that value using 'client.timing'
   var fn = function(a, b) { return a + b };
-  var metricsArgs = { 'stat': 'response_time' };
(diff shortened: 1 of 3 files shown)


ericapisani:
Looking over these changes again, I realized that they aren't achieving what I'm intending them to achieve. Specifically, I intended [this line](https://github.com/brightcove/hot-shots/pull/59/files#diff-cfdf1e0afcbe7fb32697031bec3cc7deR124) to be a substitute for the spread operator in ES6, but this is incorrect.

I'm closing this PR while I work to fix this, but hope that the core idea is still of interest.

PR #60, opened 2018-02-22
Add timer decorator function

ericapisani:
If you use the `timing` function to send stats such as how long a function or external request takes to execute, you likely are manually starting and stopping a timer using something like `process.hrtime()` before sending that time on to your StatsD server.

This decorator function abstracts that work away and takes care of tracking execution time and sending that info on to your StatsD server using the `timing` function for you.

Sample use of `timer`:

```Javascript
function add(a, b) {
  return a + b;
}

// Would send the time taken to execute 'add' as the 'time'
// argument to the 'timing' function
timer(add, 'add_execution_time')(2, 3);

```

Before this change, this behaviour would be achieved with something like the following:

```Javascript
var start = process.hrtime();
try {
 return add(2, 3);
} finally {
  var duration = process.hrtime(start)[1] / 1000000;
  metricsClient.timing('add-execute-time', duration, sampleRate, tags)
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ The check method has the following API:
   // parameter takes to execute (in milliseconds) and then sends that value
   // using 'client.timing'.
   // The parameters after the first one (in this case 'fn')
-  // match those in timer.
+  // match those in 'client.timing'.
   var fn = function(a, b) { return a + b };
   client.timer(fn, 'fn_execution_time')(2, 2);
 
--- a/README.md
+++ b/README.md
@@ -88,8 +88,11 @@ The check method has the following API:
   // Timing: sends a timing command with the specified milliseconds
   client.timing('response_time', 42);
 
-  // Timer: Records how long a given function takes to execute (in milliseconds),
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -88,6 +88,11 @@ The check method has the following API:
   // Timing: sends a timing command with the specified milliseconds
   client.timing('response_time', 42);
 
+  // Timer: Records how long a given function takes to execute (in milliseconds),
+  // and then sends that value using 'client.timing'
```
bdeitte: Can you make it more explicit that timer returns a function, and the parameters?  How about: 

// Timer: Returns a function that you call to record how long the first parameter takes to 
// execute (in milliseconds) and then sends that value using 'client.timing'.  The parameters
// after the first one match those in timer.

bdeitte:
Makes sense to me.  One small comment and I can merge

README.md
ericapisani: Sounds good, will update 👍

PR #61, opened 2018-04-03
Added support for using default route

tanelso2:
This is useful for dynamically setting the host in containers.
For example, most Kubernetes nodes will have a statsd server on them,
and using this will allow containers to send metrics directly to that
without extra configuration.

(Inspired by the same option in the datadogpy library)

Tested by:
    Checking the output of the getDefaultRoute() function


bdeitte:
Looks great, thanks!  I wish we could have some tests for this, but that's really the fault of this project not having any mocking lib in place.  I'll open an issue for this to be tackled later.

Issue #62, opened 2018-04-07
Split up the big file


bdeitte:
lib/statsd.js is closing in on 1000 lines.  It's overdue to split this up.

PR #67, opened 2018-05-29
Split up single file, add code coverage capabilities

drewen:
screen shot 2018-05-28 at 10 23 58 pm

#62 
* moves non-client-specific functions to `helpers.js`, and stat-specific functions (ie. `increment`, `set`, `timer`) to `statsFunctions.js` to mentally separate concerns and reduce single-file syndrome
* Adds nyc to create coverage reports

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 node_modules
 .idea/
 *.iml
+.nyc_output
+coverage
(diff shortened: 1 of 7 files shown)


bdeitte:
Thank you @drewen :) I will look at this tonight!

bdeitte:
Looks great, thanks.  Making it a minor change shortly (while technically a patch just to be on the safer side with the number of updates).

PR #68, opened 2018-05-30
Typescript declaration for the 'timer' method

emou:
Tested with a project written in typescript by locally linking the `hot-shots` module with this change applied and issuing metrics with a call like this:

```
this.metrics.timer(this.someMethod.bind(this), 'someMethodStats')(someObject)
```


bdeitte:
Looks good, thanks

Issue #43, opened 2017-06-13
Methods lose parameters when given empty object for sampleRate


bdeitte:
This was seen from someone another team I work with, and entering to be fixed later and noted in case others run into it.  The case ran into was like this:

```
hotShots.timing(event, milliseconds, {}, tags);
```

In that case, because an empty object is given for sampleRate, the sendAll() code believes this is a callback.  So tags or any other parameters after this are lost.  This is odd behavior having to do with the legacy allowing for shortening of parameters in the code in different ways.  To fix this and keep the behavior, a smarter check needs to be put in place when checking to see if sampleRate really is the callback by also checking later parameters to see if they are there.  And/or send a warning log on this- console.log should be used more in the code for cases where things are clearly off.

PR #69, opened 2018-06-02
Add overload types for stats functions

drewen:
Stats functions provided by hot-shots allow certain parameters to be safely omitted without mistaking them for others. A common example is omitting the `sampleRate` parameter, but still using `tags` and/or a `callback` function. Reviewing those functions, that looks to be the case (#43 aside)

As written, the types allow for omitting certain parameters, but only if all parameters before it are included. The example above of omitting `sampleRate`, but still using `tags` was not properly expressed by the existing types.

This PR adds overload types for the stat collection functions to reflect their actual capabilities, tested by replacing the types file in a separate typescript project and removing parameters with no typescript errors reported.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -61,29 +61,62 @@ declare module "hot-shots" {
   }
 
   export type StatsCb = (error: Error | undefined, bytes: any) => void;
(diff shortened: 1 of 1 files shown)


bdeitte:
Looks good, and thanks again!

PR #71, opened 2018-06-08
Add support for async functions for client.timer

### LINKED ISSUE #66: [Feature Request] support async functions for client.timer
Currently, client.timer works only for synchronous code. I think it would be nice to allow the input function to return a promise and then run the metrics after the promise resolves. This would also allow for async functions to me measured using the `client.timer` API.

If you think this is a good idea, I'd be happy to put up a PR for it.

DavidWittman:
Run `await func` to resolve any promises before sending the timing metric.

I'm a little concerned that these changes may break any instances of clients chaining promises with the timer method, but those timing stats would be incorrect anyway since they don't wait for the promise to resolve. I'm open to suggestions for how to support promise chaining while still capturing the right timing.

This also unfortunately causes some linting errors because JSHint doesn't seem to support async/await yet. I added some comments to ignore linting on the lines with async/await calls.

Closes #66

### DIFF
--- a/lib/statsFunctions.js
+++ b/lib/statsFunctions.js
@@ -26,10 +26,10 @@ function applyStatsFns (Client) {
   Client.prototype.timer = function (func, stat, sampleRate, tags, callback) {
     var _this = this;
 
-    return function () {
+    return async function () {
       var start = process.hrtime();
       try {
-        return func.apply(null, arguments);
+        return await func.apply(null, arguments);
       } finally {
         // get duration in milliseconds
         var duration = process.hrtime(start)[1] / 1000000;
--- a/lib/statsFunctions.js
+++ b/lib/statsFunctions.js
@@ -26,10 +26,10 @@ function applyStatsFns (Client) {
   Client.prototype.timer = function (func, stat, sampleRate, tags, callback) {
     var _this = this;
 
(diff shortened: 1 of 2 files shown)


bdeitte:
@DavidWittman Thanks for the PR!  We have a very wide group of people using this library, and I'm not willing to say that Node 6 (or even Node 4) isn't supported here.  Given async fails those builds, as shown above, I can't accept this PR until it's done in a way where this passes.

bdeitte:
I don't unfortunately- it might be possible but can't think of a way myself to approach this.

brandondoran:
@DavidWittman @bdeitte I'm interested in this async timer as well. `async`/`await` wouldn't work natively prior to Node 8. You could introduce a build step using babel or Typescript and target Node 4, but that adds complication and it may be easier to use Promises.  Would probably need to check that the function looks like a Promise or not before calling `.then` or `.catch`.  Another option would be to add a new `timerAsync` and leave the existing `timer` alone.

Also, I think you'd want to not swallow errors. Whether or not you report timing on errors, I'm not sure.

bdeitte:
Closing for now.  Please feel free to open up with someone that passes the tests if you have any ideas.  Thanks.

PR #73, opened 2018-06-14
Fix `timer` function in `lib/statsFunctions.js`…

mjesuele:
… to properly interpret `process.hrtime`

According to the [node documentation](https://nodejs.org/api/process.html#process_process_hrtime_time), the concluding call to `process.hrtime`:
>returns the current high-resolution real time in a `[seconds, nanoseconds]` tuple `Array`, where `nanoseconds` is the remaining part of the real time that can't be represented in second precision.

The `timer` function was previously throwing away the seconds part; this PR includes them.


bdeitte:
Oh yikes, definitely good to fix!

PR #78, opened 2018-07-04
Add support for recording buffers in mock mode

### LINKED ISSUE #77: Keep buffers from mocked calls for test inspection.
We're trying to write tests for our stats client which wraps hotshots. There are some instances where we can't supply a callback to each 'count' or 'timing' call. In order to test that our stats are emitting properly, we're having to use https://github.com/mattrobenolt/node-mock-udp to ensure that we're sending the right data.

It would be nice if hot-shots would record the outgoing buffers when `{mock: true}` is passed.

Willyham:
Fixes #77

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -64,6 +64,11 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   this.bufferHolder = options.isChild ? options.bufferHolder : { buffer: "" };
   this.errorHandler = options.errorHandler;
 
+  // If we're mocking the client, create a buffer to record the outgoing calls.
+  if (this.mock) {
+    this.mockBuffer = [];
+  }
+
   // We only want a single flush event per parent and all its child clients
   if(!options.isChild && this.maxBufferSize > 0) {
     this.intervalHandle = setInterval(this.onBufferFlushInterval.bind(this), this.bufferFlushInterval);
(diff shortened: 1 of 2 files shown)


bdeitte:
Looks good, thanks.  I'm going to make a doc update as part of this and will then make a release with it.

sdeprez:
👋 please be aware that this feature caused OOM on our servers because of the ever-growing array. We should be able to opt-out of this feature, or add a limit on the size / auto-deleting the old ones. Not sure which solution is the best, but at least a note in the docs would be helpful because otherwise it's easy to not think about it. Thanks!

Willyham:
Hey @sdeprez, this array should only grow when `mock:true` is passed (default is false). That mode is only expected to be used in tests rather than in production. Can you elaborate a little bit on your use-case?

sdeprez:
Hey thanks for the quick reply. Our use-case is that we have different environments, all of them in prod, but for some of them we don't to push any metrics (to simplify, they're deployed into someone else infrastructure and we don't want to push metrics from there). But the growing buffer, the `mock` would have been a very easy solution for that.

Now I understand that it's quite a specific situation, so just a note in the README.md could be an acceptable solution!

Willyham:
Makes sense. I think a note is a no-brainer (I'll draft something). Other potential solutions are to have a fixed size array like you suggest, or we add a `silent` mode which is like `mock` but nothing is recorded.

Issue #76, opened 2018-07-02
Add support for TCP connections


honzajavorek:
Thanks for this library! It took me a while to stumble upon it (not really intuitive name, I guess all have been taken), but the greater is the joy a maintained statsd library exists! And it even mentions Telegraf, which I'd like to use ❤️ 

The only thing I'm missing is a TCP connection support as laid out in https://github.com/sivy/node-statsd/pull/67. What would be the best way to help the TCP support to arrive?

I could obviously try to copy-paste @remie's PR to this repository, keeping him as the author of the commits, if that's the only thing missing and he agrees with it...

bdeitte:
If you brought over that PR, that sounds great.  I don't think copy-pasting the code over will work exactly, as there has been some splitting of the one main file into more than one here.  But it shouldn't be too much extra work.  While I am not actively developing in this project, I'm quick to review and merge things in.

And you are right as mentioned in that other PR- this hasn't brought into this project yet.  There is another related outstanding issue as well with https://github.com/brightcove/hot-shots/issues/53

PR #79, opened 2018-07-09
Add TCP support

michalholasek:
This PR builds upon @remie's great work at https://github.com/sivy/node-statsd/pull/67, with couple shortcomings:
* I was not able to resolve problems with global tags and `should properly send a and b with the same value` test cases (skipped with `xit`). @bdeitte, could you please point me to correct direction to what is missing / wrong?
* TCP test cases are pretty much copy/paste from the PR mentioned above - since there are lot of changes to this codebase since, it is possible that some test cases are missing

/cc @honzajavorek


bdeitte:
Thanks for this, great to see!  A few things:
- I see there's a conflicting file- looks like a merge from master resolve needs to happen here?
- I don't like the the idea of having so much duplication in tests.  This is very likely to get out of sync and be problematic for further updates.  Is there a way this can be done where both ways to test share with each other?
- Could you tell me how you are seeing some of these tests fail?  I can take a look at another time in more depth but can't do that at the moment, but if you can let me know what you're seeing now I can see what I can do to help out.
Thanks again!

lib/statsd.js
```diff
@@ -42,18 +43,28 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
       maxBufferSize : maxBufferSize,
       bufferFlushInterval: bufferFlushInterval,
       telegraf    : telegraf,
-      sampleRate    : sampleRate
+      sampleRate  : sampleRate,
+      protocol    : protocol
     };
   }
 
+  function createSocket(args) {
```
honzajavorek: I'd use `options` instead of `args`, but both are saying nothing about the contents, so it's just a convention nitpick.

lib/statsd.js
```diff
@@ -91,6 +102,10 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
     global.statsd = this;
   }
 
+  if (options.protocol === 'tcp') {
+    this.socket.setKeepAlive(true);
+  }
```
honzajavorek: Shouldn't this be a part of the `createSocket` "method"?

honzajavorek:
Also, would you mind removing the `xit` so we can see (after the conflict is resolved) what exactly fails here in the CI? The Travis is testing ~4 versions of Node here, so maybe we can factor out whether the issue isn't specific to a particular Node version?

honzajavorek:
When running locally on Node 8, I'm getting following failures randomly (flakily) even when I keep the `xit` marks intact:

```
1) StatsD [TCP] #globalTags should not add global tags if they are not specified:
     Uncaught Error: connect ECONNREFUSED 127.0.0.1:8125
      at Object._errnoException (util.js:992:11)
      at _exceptionWithHostPort (util.js:1014:20)
      at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1186:14)

  2) StatsD [TCP] #globalTags Uncaught error outside test suite:
     Uncaught TypeError: Cannot read property 'currentRetry' of undefined
```

They occur especially when running the whole test suite repeatedly. Repeated `--grep=TCP` gets me into state when I get the TCP tests with `xit`s passing pretty consistently. Strange!

michalholasek:
The conflict was probably caused by `test_statsd.js -> test_statsd_udp.js` rename - I removed the commits with changes in `test_statsd.js` for now. `xit` test cases are also unskipped.

lib/statsd.js
michalholasek: Addressed in https://github.com/brightcove/hot-shots/pull/79/commits/69de8cbfcad688e76ea621dcb5303250170f3c09.

honzajavorek:
> race conditions in create server -> assert -> close server logic

I suspect that as well. We should hunt those down.

michalholasek:
@honzajavorek I do. I am not sure it will be possible without the rewrite 😞

honzajavorek:
My hunch is the `tcpTest` function does something incorrectly and makes all the tests non-deterministic.

michalholasek:
@honzajavorek Changes made in https://github.com/brightcove/hot-shots/pull/79/commits/b571d8cdaaa5c9557184cb9ae3d64461fe547514 ensure that the tests pass/fail consistently now. The solution is not ideal, but at least we can now focus on making them pass.

test/test_statsd_tcp.js
```diff
@@ -5,666 +5,628 @@ var net = require('net');
...
+          parts = packet.split("\n");
         } else {
-          metrics = [packet];
+          parts = [packet];
         }
 
-        metrics.forEach(function (metric) {
-          test(metric, server);
-        });
+        var metrics = parts.filter(part => part !== '');
```
honzajavorek: test/test_statsd_tcp.js: line 37, col 41, 'arrow function syntax (=>)' is only available in ES6 (use 'esversion: 6').

test/test_statsd_tcp.js
```diff
@@ -5,666 +5,628 @@ var net = require('net');
...
-      var statsd = new StatsD({ protocol: 'tcp' });
-      assert.ok(statsd.socket instanceof net.Socket);
+      createServer(function (address) {
+        statsd = new StatsD({
+          host: address.address, 
+          port: address.port,
+          protocol: 'tcp'
+        });
+        assert.ok(statsd.socket instanceof net.Socket);
+        done();
```
honzajavorek: test/test_statsd_tcp.js: line 72, col 9, 'done' is not defined.

honzajavorek:
BTW I can still see `#increment` being strangely nested inside `#gauge` in the tests output, which is rather suspicious.

```
#gauge
      ✓ should send proper gauge format without prefix, suffix, sampling and callback
      ✓ should send proper gauge format with tags
      ✓ should send proper gauge format with prefix, suffix, sampling and callback
      4) should properly send a and b with the same value
      5) "after each" hook
      #increment
```

I'll try to help with debugging this. Maybe I can figure something out. @bdeitte if you don't have any immediate ideas on what could be wrong with either the TCP implementation or the tests, then feel free to ignore it here until we get the thing passing. And sorry for the noise.

michalholasek:
@bdeitte The build is green now, could you please take another look? Thank you.

bdeitte:
@michalholasek Thanks for getting the tests working and doing all this work.  Unfortunately I still have the same feeling I have before- the duplication of tests in the way they are isn't something I am comfortable with.  The tests are already drifting apart from what I see, and this is only going to increase without careful diligence over time.  There needs to be something done to get them closer in line, even if it's not exactly right. I know that you mention this could be done in a followup PR, but that is not a great place for this codebase to be in.  If you need to have this right away and can't make those changes, I would suggest using your branch for now until that can be done.  (Although I hope you will make them now, would be great to have this in!)

I thought about this some more, to try to think of something that isn't too painful to do but which gets more at what I'm worried about here.  Here's my suggestion: https://gist.github.com/bdeitte/4ae7fd92edb490e2adcd22e38d62a5a6

What I did in the above is take your new tests, rip out everything other than globalTags tests, and then try to add in the UDP tests.  I ran this in your branch and it passes there.

If you want to try to add this way of doing things, here's you would need to do:
1. Start with test_statsd_udp.js, and rename createServer to createTCPServer.  Change all references inline.  Also add in createUDPServer as shown above.
2. Copy each "describe" section you've made and paste a new one below it.  Name one UDP and one TCP.  For the UDP one, switch all TCP references to UDP
3. Run it all and hopefully see it all pass
4. Look back in test_statsd.js to see if anything is missing in there- I know there's some tests that were just in there.  Copy them in and switch to createTCPServer()
5. Rename test_statsd_udp.js to test_statsd.js

In the future, it would be good to split these tests into multiple files (while still keeping UDP/TCP tests for related functionality together) as this file is really big.  It'd also probably be good to think more about ways to shorten some of the boilerplate in the tests.  But I started to think about that and realized it's not what I'm asking here and definitely don't want to make this something more than it is, which is just a request to have similar tests live by each other for easier maintenance.  Thanks!

lib/statsd.js
```diff
@@ -216,6 +239,11 @@ Client.prototype.send = function (message, tags, callback) {
       message += '|#' + mergedTags.join(',');
     }
   }
+
+  if (this.protocol === 'tcp') {
+    message += '\n';
+  }
```
honzajavorek: I think this should be in the `sendMessage` method as that one deals with the quirks of the individual transmission protocols. This is an implementation details of sending the message over TCP and IMHO should be encapsulated by `sendMessage`.

lib/statsd.js
```diff
@@ -42,18 +43,37 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
...
+      socket = dgram.createSocket('udp4');
+    }
+
+    return socket;
+  }
+
   // hidden global_tags option for backwards compatibility
   options.globalTags = options.globalTags || options.global_tags;
 
+  this.protocol = options.protocol;
```
honzajavorek: Nitpick: Although it's against various code styles, I'd stick to the code style already in place and add the alignment whitespaces:

```
- this.protocol = options.protocol;
+ this.protocol    = options.protocol;
```

honzajavorek:
Just for the record, once this is done, it closes https://github.com/brightcove/hot-shots/issues/76 (I miss the connection when browsing the issues/PRs here 😄 )

michalholasek:
@bdeitte I decided to go with the test files split after all (note that I finished about 1/3 so far, there's a lot of them and I had to debug almost every single TCP test) - could you please just quickly skim through if it looks good enough so far? Thank you.

bdeitte:
@michalholasek This is looking great.  Also a +1 on the style updates, I have no issues there either.  Thanks for all the work.

lib/statsd.js
```diff
@@ -42,26 +43,55 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
...
+        } else {
+          console.log(errMessage);
+        }
+      }
+    } else {
+      socket = dgram.createSocket('udp4');
+    }
+
+    return socket;
+  }.bind(this);
```
bdeitte: Doing it this way means we have two ways to have functions in here.  Could you make this like the existing "_send" method?  Alternatively, if you feel strongly about it, it's not that I'm a big fan of "_send"- you could also make that method like this one.

lib/statsd.js
```diff
@@ -42,26 +43,55 @@ var Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
       maxBufferSize : maxBufferSize,
       bufferFlushInterval: bufferFlushInterval,
       telegraf    : telegraf,
-      sampleRate    : sampleRate
+      sampleRate  : sampleRate,
+      protocol    : protocol
     };
   }
```
bdeitte: Someone is going to do "TCP" at some point and get confused.  How about adding:
```
if (options.protocol) {
  options.protocol = options.toLowerCase();
}
```

package.json
```diff
@@ -31,8 +31,8 @@
   },
   "scripts": {
     "coverage": "nyc --reporter=lcov npm test",
-    "test": "mocha -R spec",
-    "lint": "jshint lib/**.js test/**.js",
+    "test": "mocha -R spec --timeout 5000 test/index.js",
+    "lint": "jshint lib/**.js",
```
bdeitte: Why not lint the tests anymore?  If you had an issue, can we just exclude that with a jshint ignore comment?

test/buffer.js
```diff
@@ -0,0 +1,184 @@
+'use strict';
+
+var assert = require('assert');
+
+var createStatsdClient = require('./helpers').createStatsdClient;
+var createTCPServer = require('./helpers').createTCPServer;
+var createUDPServer = require('./helpers').createUDPServer;
```
bdeitte: These tests and the abstractions in them look great- thanks!

package.json
michalholasek: Addressed in https://github.com/brightcove/hot-shots/pull/79/commits/34215ff69f34000493d3d897416f542837d65d1c.

lib/statsd.js
michalholasek: Addressed in https://github.com/brightcove/hot-shots/pull/79/commits/fe2d446bf926e8b5107804690e275c431b2ca3fc.

lib/statsd.js
michalholasek: I added the `bind` to appease `jshint`, which is not good enough reason, I agree. I changed the `createSocket` function signature to accept instance argument instead (https://github.com/brightcove/hot-shots/pull/79/commits/dee23b3c6c80921182133127995ac74caf7d5bc4).

michalholasek:
@bdeitte Thank you for your time you've put into feedback/reviewing! @remie Thank you for the original work, it helped a lot 🙂

PR #80, opened 2018-07-26
 fixing maths error on nanosecconds -> microseconds

chrismatheson:
It took me quite a while to figure out what going on because i assumed my code was wrong before digging into hot-shots, but i think there is a math error with the `timer` convenience function.

So ive added a test which (i beleive) highlights the problem, and then a fix in a separate commit (very simple)

Hope this is an easy one to review & merge as id really love not to run a fork in production :)


chrismatheson:
closing so i can re-open with the correct branch from my own fork :)

PR #81, opened 2018-07-26
asyncTimer

chrismatheson:
introduction of function which understands async return and instruments appropriately 

should resolve https://github.com/brightcove/hot-shots/issues/66

### DIFF
--- a/test/timer.js
+++ b/test/timer.js
@@ -1,5 +1,5 @@
 'use strict';
-
+var StatsD = require('../lib/statsd');
 var assert = require('assert');
 
 var createStatsdClient = require('./helpers').createStatsdClient;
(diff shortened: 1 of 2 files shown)


bdeitte:
I see that the Node 8 says they failed, but I just reran it and it's fine.  I'll take an actual look at this PR (and your other one) for merging later tonight.  Thanks!

test/timer.js
```diff
@@ -105,5 +105,32 @@ module.exports = function runTimerTestSuite() {
...
+
+      return instrumented().then(function() {
+
+        var stat = statsd.mockBuffer[0];
+        var name = stat.split(/:|\|/)[0];
+        var time = stat.split(/:|\|/)[1];
+
+        assert.equal(name, 'name-thingy');
+        assert.ok(parseFloat(time) >= 100);
+        assert.ok(parseFloat(time) < 110);
```
bdeitte: I have the same worry here about creating a jittery test, perhaps a little higher?

bdeitte:
One small comment.  Thank you for all of your work here, great to see!

Heads up @DavidWittman that there's another solution that came through here.

chrismatheson:
sure thing @bdeitte 

I didn't want to step on any toes, but i was curious about the structure of the tests. Seems that the specific including of the test modules and wrapping in a top level describe is redundant if the recursive option for mocha is used.

I only noticed because my IDE wont recognise the mocha tests and let me easily run a single one etc.

Id be happy to refactor the tests a little if thats something that would be acceptable ?

bdeitte:
@chrismatheson Always happy to take improvements.  :)  The tests were recently refactored to not be in one huge mess of a file, in https://github.com/brightcove/hot-shots/pull/79.  So it hasn't been around for long in the current form and if there's improvements to be made, I'll always review and pull in.

PR #82, opened 2018-07-26
fixing maths error on nanosecconds -> microseconds

chrismatheson:
It took me quite a while to figure out what going on because i assumed my code was wrong before digging into hot-shots, but i think there is a math error with the timer convenience function.

So ive added a test which (i beleive) highlights the problem, and then a fix in a separate commit (very simple)

Hope this is an easy one to review & merge as id really love not to run a fork in production :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
       "integrity": "sha1-wHshHHyVLsH479Uad+8NHTmQopI=",
       "dev": true,
       "requires": {
-        "balanced-match": "1.0.0",
+        "balanced-match": "^1.0.0",
         "concat-map": "0.0.1"
       }
     },
@@ -27,7 +27,7 @@
       "dev": true,
       "requires": {
         "exit": "0.1.2",
-        "glob": "7.1.2"
+        "glob": "^7.1.1"
       }
     },
     "commander": {
@@ -48,7 +48,7 @@
       "integrity": "sha1-8CQcRXMKn8YyOyBtvzjtx0HQuxA=",
       "dev": true,
       "requires": {
-        "date-now": "0.1.4"
+        "date-now": "^0.1.4"
       }
     },
     "core-util-is": {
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -16,7 +16,7 @@
       "integrity": "sha1-wHshHHyVLsH479Uad+8NHTmQopI=",
       "dev": true,
       "requires": {
-        "balanced-match": "1.0.0",
+        "balanced-match": "^1.0.0",
```
bdeitte: The package-lock has a whole lot of updates to not lock to a specific version.  Not that it matters a lot, since it's all dev dependencies, but do you know why that happened?  I haven't seen this before so I'm curious.

test/timer.js
```diff
@@ -105,5 +106,28 @@ module.exports = function runTimerTestSuite() {
...
+    it('should record "real" time of function call', function () {
+      var statsd = new StatsD({mock:true});
+      var instrumented = statsd.timer(sleep(100), 'blah');
+
+      instrumented();
+
+      var timeFromStatLine = statsd.mockBuffer[0].match(/blah:(\d+\.\d+)\|/)[1];
+
+      assert.ok(timeFromStatLine >= 100);
+      assert.ok(timeFromStatLine < 120);
```
bdeitte: I worry about this becoming a jittery test with this.  How about "< 200" just in case?

test/timer.js
```diff
@@ -105,5 +106,28 @@ module.exports = function runTimerTestSuite() {
...
+    });
   });
+
+
 };
+
+
+
+
```
bdeitte: Can you remove the extra lines here?

package-lock.json
bdeitte: I think it makes sense to just leave this out, or I can do that when merged to master

bdeitte:
A couple small requests and then I can merge and make a new version.  Thanks for the catch and fix!

test/timer.js
chrismatheson: Yeah, the upper bound was simply to sanity check really so ill up the limit :)

package-lock.json
chrismatheson: errrr pass :) npm 6 + my particular settings side affect i think, ill take it out ....

PR #83, opened 2018-07-30
Add increment with tags example to readme

blimmer:
This wasn't clear to me at first because the increment value is inferred. This commit adds an explicit example so future folks don't have to dive into the source code.


bdeitte:
Thanks for the docs!

blimmer:
No problem. Thanks for this project!

PR #86, opened 2018-08-29
Fix types: default export is the `StatsD` class

mohsen1:
Node REPL:

```
> const StatsD = require('hot-shots')
undefined
> typeof StatsD
'function'
> new StatsD()
Client {
  host: 'localhost',
// ...
```

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -124,5 +124,4 @@ declare module "hot-shots" {
   }
 }
 
-declare const StatsDClient: StatsD;
-export default StatsDClient;
+export default StatsD;


mohsen1:
Per TypeScript team recommendation, you better off letting DefinitelyTyped manage your declaration files or rewrite this whole repo in TS. Maintaining declaration files is not fun if you're not a fan of TS :)

bdeitte:
It's not clear to me- is this a backwards compatible change?  @drewen thoughts here?

As for how to do types differently, this is 100% a community run project right now.  I am active in reviewing and merging things (a bit slower than usual on this one!) but otherwise everything comes from non-maintainers.  If you want to suggest and run with one of those ideas, feel free to post it in Issues and we can have a discussion on it.

bdeitte:
Closing this out, given it was not clear to me if this is backwards compatible.  Please feel free to open anytime if you can clear this up, thanks.

PR #87, opened 2018-09-13
Add asyncTimer types

etaoins:
These are based off the types of `timer` except:

- The inner and outer function return `Promise<T>` instead of a bare value

- There is no `value` parameter


bdeitte:
Looks good to me, thanks.

PR #88, opened 2018-11-06
Add mockBuffer to types.d.ts so TS tests may access for testing purposes

stieg:
Wanted to access this in TS to validate the content of the messages during
testing. Found this wasn't there. So here we are.


bdeitte:
Makes sense, thanks.

PR #89, opened 2018-11-23
Improve tests, require Node 6 or higher, and a few misc updates

bdeitte:
A whole lot of changes I've been working on that will require a major version update.  I will be letting these sit in a PR for awhile and ping some people in time to see if I can get some reviews.  If you review yourself, or some part of this, please let me know in a comment.  Thanks.

Next up, after this goes through, is that I'd like to run https://github.com/lebab/lebab on the project.  Then those two things together will be a major update that I'm excited to get out there and really upgrade this codebase.

Changes in here, not related to testing:
- Require Node 6.  Older versions may have worked before, and Node 4 was tested before.  But now I want to be able to do some newer things (and already do one).  So start supported just the versions under LTS.
- Stop using "new Buffer", that is deprecated, and use Buffer.from() instead
- Update close() to handle errors better, not doubling up in error messages and not leaving uncaught errors
- Update README to be better
- A few misc removals of things not needed

Changes in here related to testing:
- Remove a lot of duplication in tests.  Excluding package-lock.json, more than 1300 lines were removed from the project in this PR.
- Start using Mocha 4
- Stop using index.js for testing.  Move to allowing individual test files running independently
- Start using the code coverage report as part of the build
- Remove the ignoring of errors on close of tests, and tear down tests in general better
- Fix up tests here and there... I don't remember all of this as there was a ton of test work to make this better


bdeitte:
@drewen @ericapisani @michalholasek I looked back at PRs for people who have submitted PRs that were substantial, to find people to review here.  Your three names came up in the last year.  No worries if you can't review, but a heads up if you want to take a look.

PR #91, opened 2018-12-17
Handle case when value is not an number

msmnc:
Handle case, when send number as a tag value. Tag example: `{'year': 2018}`


bdeitte:
This is a regression, right?  It sounds like one to me from the big 6.0.0 update.  I'll merge now, thanks.

bdeitte:
It would also be great to have a text case on this, but wanted to merge right away in case it is affecting others.

Issue #41, opened 2017-04-13
close() call does not seem to send data before closing


sashman:
Hey,

I have the following code: 
```
    const graph =new StatsD({ host: 'example.com', port: 8125})

    const responseTime = Date.now() - this.startTime
    graph.timing('.response_time', responseTime)

    graph.close(function(err) {
      console.error('The close did not work quite right: ', err)
    });
```

With the `close` call there, no data is sent out at all. Am I using it correctly? The docs specify all data should be sent before the socket is closed.

Tested with `tcpdump -n udp dst port 8125`

bdeitte:
timing() is asynchronous.  There is a callback that you can add to it, that's optional, that returns when the data has really been sent.  If within that callback you called closed, I would expect you see the data sent.  If that's not the case, please feel free to reopen!

bdeitte:
Yes, I see there is a bug here.  Thanks for catching this!  I am not able to find the time to fix this myself, but I am happy to accept a PR from anybody who does fix this and will keep open this issue.

What needs to happen here is that the flushQueue callback can't be ignored:
https://github.com/brightcove/hot-shots/blob/master/lib/statsd.js#L526
The callback for flushQueue needs to be listened to within close() before moving on with anything.

guyisra:
@bdeitte You mean it should be 
`this.flushQueue( () => this.socket.close())` ?

bdeitte:
@guyisra Something like that, yes.  socket.close() can thrown an error so the existing catch call needs to be in there.  Also this library isn't currently using babel and has never formally deprecated any version of Node, so we'd have to do a major version bump to use the now-usual ES6 goodness.

guyisra:
From what I see is that when close() is called, the buffer is empty and the socket closes immediatly. But then the message that was sent (before the close()) is trying to get out but the socket is already closed. It would seem that the messages that need to be sent should be enqueued (or I probably misunderstand the intent of enqueue() so when the socket gets closed, it will go over the bufferHolder and finish the enqueued message before closing. Does this make sense?

bdeitte:
I think what I propose above would solve this, and you don't need any other enqueued message solution.  The buffer can be empty but the message can still not be sent- in order to know this you have to listen to the socket callback: https://github.com/brightcove/hot-shots/blob/master/lib/statsd.js#L502

This socket callback is in the end what's listen to when you add a callback listener for flushQueue().  You then can be assured all of these messages are sent and really close the socket.

[...]

PR #93, opened 2019-01-01
Fix up how close calls work

bdeitte:
This fixes long-requested https://github.com/brightcove/hot-shots/issues/41 and adds to tests for the different permutations of this case.

Also as I dug into this, I realized that the client.socket.on('error') pattern mentioned in the docs isn't the best choice, and it's better to suggest using the errorHandler so it can catch more cases.  I've updated the docs for this while not breaking backwards compatibility.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -88,7 +88,7 @@ module.exports = {
             "error",
             "unix"
         ],
-        "lines-around-comment": "error",
+        "lines-around-comment": "off",
         "lines-around-directive": "error",
         "lines-between-class-members": [
             "error",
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -204,7 +204,7 @@ module.exports = {
         "no-useless-return": "error",
         "no-var": "error",
         "no-void": "error",
-        "no-warning-comments": "error",
+        "no-warning-comments": "off",
```
bdeitte: The eslint file is new, so still turning things off that irk me while developing

guyisra:
yay

PR #95, opened 2019-01-04
Fix the timer function type definition

mbellerose:
The timer function doesn't take a `value` parameter so removing it.

https://github.com/brightcove/hot-shots/blob/master/lib/statsFunctions.js#L29


bdeitte:
👍

PR #96, opened 2019-01-08
Readme fixes

Ithildir:
Hi,
I noticed that `errorHandler` gets called every time, even when there's no error: https://github.com/brightcove/hot-shots/blob/master/lib/statsd.js#L334.

If this is by-design, it would be nice to have it clearly written in the readme.


bdeitte:
Oh that's broken.  :/  This is newly-changed behavior in 6.1.0.  I'll fix it shortly and put out 6.1.1.  If you'd like to submit this PR with just the spelling/grammatical fixes, happy to have it.  Thanks for noting this!

PR #98, opened 2019-01-11
Fix syntax in README example

JamesMGreene:
A missing `}` had me confused for a minute as to why `errorHandler` would be passed to the `globalTags` option. 😄


bdeitte:
Well that's what I get for writing example code without trying it out.  :) . Thanks for the fix!

Issue #114, opened 2019-07-11
UDS connection doesn't repair after agent restart


danopia:
I've been testing UDS for Kubernetes applications and have noticed that restarting the agent causes the metrics calls to throw. The library doesn't seem to repair - my code is just throwing repeatedly until I restart the nodejs application.

The first call after agent restart looked like this:

```
Error: Error sending hot-shots message: Error: send -111
Error: Error sending hot-shots message: Error: Error sending hot-shots message: Error: send -111
```

After that, the rest of the attempts look like this:

```
Error: Error sending hot-shots message: Error: send -107
Error: Error sending hot-shots message: Error: Error sending hot-shots message: Error: send -107
```

Ideally these errors can be handled invisibly by reconnecting, not sure what the best course of action is here. [Datadog docs say](https://docs.datadoghq.com/developers/dogstatsd/unix_socket/#how-it-works) "When the Agent restarts, the existing socket is deleted and replaced by a new one. Client libraries detect this change and connect seamlessly to the new socket."
I used `.event()` to report JS errors, so that failed too, giving me a stacktrace:

```
(node:1) UnhandledPromiseRejectionWarning: Error: Error sending hot-shots message: Error: Error sending hot-shots message: Error: send -107
    at handleCallback (/src/node_modules/hot-shots/lib/statsd.js:369:32)
    at Client.sendMessage (/src/node_modules/hot-shots/lib/statsd.js:397:5)
    at Client._send (/src/node_modules/hot-shots/lib/statsd.js:308:12)
    at Client.send (/src/node_modules/hot-shots/lib/statsd.js:285:8)
    at Client.event (/src/node_modules/hot-shots/lib/statsFunctions.js:308:10)
    at Timeout.loop [as _onTimeout] (/src/app.js:69:9)
    at processTicksAndRejections (internal/process/task_queues.js:89:5)
```

danopia:
After reading through https://github.com/DataDog/datadog-agent/wiki/Unix-Domain-Sockets-support#implementing-uds-support-for-statsd-client-libraries I see some key points:

* the library should act like UDP by default, aka, it should not throw or block for more than 1ms when emitting metrics to UDS
* unless a 'strict' config is enabled, dropping metrics during a downed agent connection is fine
* the library should try reconnecting regularly (but not too often) after getting disconnected

Since this library isn't directly owned by Datadog, I'm not sure how motivated you are to follow Datadog's spec...

bdeitte:
[...]

PR #116, opened 2019-07-16
add options.udsGracefulErrorHandling to enable sockere-creation with …

NinjaBanjo:
…a time limit

An attempt at solving the problem mentioned in brightcove/hot-shots#114

things to note:
- this solution ignores non-uds protocols
- it is off by default
- it adds two new options
  - udsGracefulErrorHandling (boolean) enable/disable this new func (default = disable)
  - udsGracefulRestartRateLimit (int) time (ms) between creating a new socket (prevents wild thrashing)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ Parameters (specified as one object passed into hot-shots):
 * `useDefaultRoute`: Use the default interface on a Linux system. Useful when running in containers
 * `protocol`: Use `tcp` option for TCP protocol, or `uds` for the Unix Domain Socket protocol. Defaults to UDP otherwise
 * `path`: Used only when the protocol is `uds`. Defaults to `/var/run/datadog/dsd.socket`.
+* `udsGracefulErrorHandling`: Used only when the protocol is `uds`. Boolean indicating wether to handle socket errors gracefully. defaults falsey
+* `udsGracefulRestartRateLimit`: Used only when the protocol is `uds`. Time (ms) between re-creating the socket. Defaults to `1000`.
 
 ### StatsD methods
 All StatsD methods other than `event`, `close`, and `check` have the same API:
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -154,6 +154,45 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
     global.statsd = this;
   }
 
+  // only for uds (options.protocol uds)
+  // enabled with the extra flag options.udsGracefulErrorHandling
+  // will gracefully (attempt) to re-open the socket with a small delay
+  // options.udsGracefulRestartRateLimit is the minimum time (ms) between creating sockets
+  // does not support options.isChild (how to re-create a socket you didn't create?)
+  if (!options.isChild && options.protocol === 'uds' && options.udsGracefulErrorHandling) {
+    const socketCreateLimit = options.udsGracefulRestartRateLimit || 1000; // only recreate once per second
+    const lastSocketCreateTime = Date.now();
+    this.socket.on('error', (err) => {
+      const code = err.code;
+      switch (code) {
+        case 107:
+        case 111: {
+          if (Date.now() - lastSocketCreateTime >= socketCreateLimit) {
+            // recreate the socket, but only once per 30 seconds
+            if (this.errorHandler) {
+              this.socket.removeListener('error', this.errorHandler);
+            }
+            this.socket.close();
+            this.socket = createSocket(this, {
+              host: this.host,
+              path: options.path,
+              port: this.port,
+              protocol: this.protocol
+            });
+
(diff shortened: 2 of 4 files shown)


NinjaBanjo:
I had a number of issues with references in the mocha tests, so i was forced to deviate from the patterns used in other tests slightly. i tried to keep it as closely aligned as possible.

If you have any suggestions for a cleaner solution I'm open to a ideas

README.md
```diff
@@ -45,6 +45,8 @@ Parameters (specified as one object passed into hot-shots):
 * `useDefaultRoute`: Use the default interface on a Linux system. Useful when running in containers
 * `protocol`: Use `tcp` option for TCP protocol, or `uds` for the Unix Domain Socket protocol. Defaults to UDP otherwise
 * `path`: Used only when the protocol is `uds`. Defaults to `/var/run/datadog/dsd.socket`.
+* `udsGracefulErrorHandling`: Used only when the protocol is `uds`. Boolean indicating wether to handle socket errors gracefully. defaults falsey
```
bdeitte:
```suggestion
* `udsGracefulErrorHandling`: Used only when the protocol is `uds`. Boolean indicating whether to handle socket errors gracefully.  Defaults to true.
```

Note the suggestion above to default to true- it sounds like this is really broken without this being enabled, correct?  Any problems with having it default to true

lib/statsd.js
```diff
@@ -154,6 +154,45 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
     global.statsd = this;
   }
 
+  // only for uds (options.protocol uds)
+  // enabled with the extra flag options.udsGracefulErrorHandling
+  // will gracefully (attempt) to re-open the socket with a small delay
+  // options.udsGracefulRestartRateLimit is the minimum time (ms) between creating sockets
+  // does not support options.isChild (how to re-create a socket you didn't create?)
+  if (!options.isChild && options.protocol === 'uds' && options.udsGracefulErrorHandling) {
+    const socketCreateLimit = options.udsGracefulRestartRateLimit || 1000; // only recreate once per second
```
bdeitte: Could you make 1000 a constant at the top of the file?  We don't do this everywhere but trying to get better with it for cases like this.

lib/statsd.js
```diff
@@ -154,6 +154,45 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
...
+            this.socket = createSocket(this, {
+              host: this.host,
+              path: options.path,
+              port: this.port,
+              protocol: this.protocol
+            });
+
+            if (this.errorHandler) {
+              this.socket.on('error', this.errorHandler);
+            }
```
bdeitte: Could you console.log a message if the errorHandler isn't there?

bdeitte:
Understandable to me that you have to deviate from the tests in the case, don't see a problem with it.  Have some small suggestions.  Thanks for getting to this!

NinjaBanjo:
I appreciate the feedback. I'm more than happy to incorporate.

After some testing in our own environments using this code, it's not sufficient. Though the direction here is sound (i think?)

using `this.socket.on('error'` doesn't catch all the errors, as `sendMessage` swallows some if you have an `errorHandler`

I'm going to be taking a second pass at this, but it might take me a day or two.

Also, the codes `107` and `111` seem to be unique to our environment, and how we are using the datadog agent? Haven't nailed this part down yet.

The final solution should handle more error codes than 107 and 111 probably?

I would suggest this PR not be merged yet, but I would like to leave it here for anyone else that might make a pass at this before I get a chance to revisit.

bdeitte:
Thanks for the details, and good to know.  That's interesting to hear on the error codes- I have no idea on that part myself.

Enrico2:
Very much looking forward to this PR merging and released.

NinjaBanjo:
This fell off my priorities list, unfortunately. I will circle back around eventually if someone else doesn't solve this. But I can't say when.

lib/statsd.js
```diff
@@ -154,6 +154,45 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
...
+            if (this.errorHandler) {
+              this.socket.removeListener('error', this.errorHandler);
+            }
+            this.socket.close();
+            this.socket = createSocket(this, {
+              host: this.host,
+              path: options.path,
+              port: this.port,
+              protocol: this.protocol
+            });
```
msiebuhr: `createSocket(instance, args)` w. `args.protocol = 'uds'` only looks at `args.path`. Why then pass `host` and `path`?

bdeitte:
Closing in favor of https://github.com/brightcove/hot-shots/pull/123 so that hopefully we can get this merged in.  Certainly if you find you get back to this and want this reopened (if the other PR does end up getting stalled), I more than welcome it!

PR #118, opened 2019-09-09
Fix cacheDns option when obtaining host from DD_AGENT_HOST

jfirebaugh:
Look up `this.host`, not `options.host`, so that caching works when the host is obtained from the `DD_AGENT_HOST` environment variable or the default value of `"localhost"` is used. `this.host` is initialized with `options.host || process.env.DD_AGENT_HOST || 'localhost'`.


bdeitte:
Clear bug, makes sense, thanks!

PR #123, opened 2019-09-26
Add udsGracefulErrorHandling

msiebuhr:
Tried fixing the feedback from #116 

To be totally correct, it probably should've been a PR against NinjaBanjo/hot-shots, but I figured this might move a bit quicker (otherwise: feel free to cherry-pick my changes into #116)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ Parameters (specified as one object passed into hot-shots):
 * `useDefaultRoute`: Use the default interface on a Linux system. Useful when running in containers
 * `protocol`: Use `tcp` option for TCP protocol, or `uds` for the Unix Domain Socket protocol. Defaults to UDP otherwise
 * `path`: Used only when the protocol is `uds`. Defaults to `/var/run/datadog/dsd.socket`.
+* `udsGracefulErrorHandling`: Used only when the protocol is `uds`. Boolean indicating wether to handle socket errors gracefully. defaults falsey
+* `udsGracefulRestartRateLimit`: Used only when the protocol is `uds`. Time (ms) between re-creating the socket. Defaults to `1000`.
 
 ### StatsD methods
 All StatsD methods other than `event`, `close`, and `check` have the same API:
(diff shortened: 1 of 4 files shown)


bdeitte:
I don't see it as a problem at all myself to just use this PR, prefer that as well to have this merged in faster.

bdeitte:
@msiebuhr Looks like the inline feedback is taken care of, thanks!  I do though see a comment from @NinjaBanjo at the top level which I don't see changes for.  There's two things they bring up:

1. After some testing in our own environments using this code, it's not sufficient. Though the direction here is sound (i think?).   using this.socket.on('error' doesn't catch all the errors, as sendMessage swallows some if you have an errorHandler
2. Also, the codes 107 and 111 seem to be unique to our environment, and how we are using the datadog agent? Haven't nailed this part down yet.  The final solution should handle more error codes than 107 and 111 probably?

Is this something you are now running, and does it work fine for you?  I'm fine if that is the case, and being iterative on this going in.  But the second comment in particular is curious- are those really unique per env, do you know?

alessio-modus:
My team is trying integrate with Datadog using the `UDS` protocol.
I don't have the bandwidth to contribute code at the moment but I took the liberty of compiling a list of error codes along with retry behavior. Hope this helps 😄 

| Name | Code | Description | Retry |
|---|---|---|---|
ENETDOWN | 100 | Network is down | ✅ 
ENETUNREACH | 101 |	Network is unreachable | ✅ 
ENETRESET |	102	| Network dropped connection on reset | ✅ 
ECONNABORTED | 103 | Software caused connection abort | ❌ 
ECONNRESET | 104 | Connection reset by peer | ✅ 
ENOBUFS | 105 | No buffer space available | ❌ 
EISCONN | 106 | Socket is already connected | ❌ 
ENOTCONN | 107 | Socket is not connected | ✅ 
ESHUTDOWN | 108 | Can't send after socket shutdown | ❌ 
ETOOMANYREFS | 109 |Too many references: can't splice | ❌ 
ETIMEDOUT | 110 | Connection timed out  | ✅ 
ECONNREFUSED | 111 | Connection refused | ✅ 

I also verified that 107 and 111 should be retried by testing with Datadog agent hosted on Kubernetes.

bdeitte:
Thanks for all of the details @alessio-modus!  Ok here's my plan:

1. Merge this in tomorrow, assuming I don't hear any objections or changes coming from @alessio-modus @msiebuhr or @NinjaBanjo.  It has enough done it, and it was broken enough before, to get this in.  
2. Create an issue for the codes that are mentioned above, to potentially add to what is done here
3. Ignore the comment from @NinjaBanjo for now on "using this.socket.on('error' doesn't catch all the errors, as sendMessage swallows some if you have an errorHandler".  If I hear back there, can be a separate issue to be addressed later.

PR #124, opened 2019-09-26
Fix crasher when closing Unix Datagram Sockets without callback

msiebuhr:
```javascript
const statsd = new hotShots.Client({type: 'uds'});
uds.close(function () {}); // works fine
uds.close(); // TypeError: callback is not a function
```

Elsewhere in `close()`, the existence of callback is checked a lot, so I assume this to be an oversight.


bdeitte:
Definitely just an oversight, thanks!

PR #129, opened 2019-10-08
Abstract transports

runk:
.. was looking into adding another transport (raw stream), however the way different protocols were managed was quite hectic. Doing abstraction as a first step toward that.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,3 +1,0 @@
-{
-  "singleQuote": true
-}
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "singleQuote": true
+}
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1,0 +1,6 @@
+exports.PROTOCOL = {
+  TCP: 'tcp',
+  UDS: 'uds',
+  UDP: 'udp',
+  STREAM: 'stream'
+};
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -352,10 +352,6 @@ Client.prototype.sendMessage = function (message, callback) {
     return callback(new Error('No transport set. Check previous errors for details.'));
   }
 
-  if (this.protocol === PROTOCOL.TCP && message.lastIndexOf('\n') !== message.length - 1) {
-    message += '\n';
-  }
-
   const handleCallback = (err) => {
     this.messagesInFlight--;
     const errFormatted = err ? new Error(`Error sending hot-shots message: ${err}`) : null;
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -352,6 +352,10 @@ Client.prototype.sendMessage = function (message, callback) {
     return;
   }
 
+  if (!this.socket) {
+    return callback(new Error('No transport set. Check previous errors for details.'));
+  }
+
   if (this.protocol === TCP && message.lastIndexOf('\n') !== message.length - 1) {
     message += '\n';
   }
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -383,7 +383,7 @@ Client.prototype.sendMessage = function (message, callback) {
   try {
     this.messagesInFlight++;
     if (this.protocol === TCP) {
-      this.socket.write(buf, 'ascii', handleCallback);
+      this.socket.send(buf, handleCallback);
     } else if (this.protocol === UDS) {
       this.socket.send(buf, handleCallback);
     } else {
@@ -482,7 +482,7 @@ Client.prototype._close = function (callback) {
 
   try {
     if (this.protocol === TCP) {
-      this.socket.destroy();
+      this.socket.close();
     } else if (this.protocol === UDS) {
       this.socket.close();
       // this close is synchronous, and the socket will not emit a
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -349,7 +349,7 @@ Client.prototype.sendMessage = function (message, callback) {
   }
 
   if (!this.socket) {
-    return callback(new Error('No transport set. Check previous errors for details.'));
+    return callback(new Error('No protocol set. Check previous errors for details.'));
   }
 
   const handleCallback = (err) => {
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -387,7 +387,8 @@ Client.prototype.sendMessage = function (message, callback) {
     } else if (this.protocol === UDS) {
       this.socket.send(buf, handleCallback);
     } else {
-      this.socket.send(buf, 0, buf.length, this.port, this.host, handleCallback);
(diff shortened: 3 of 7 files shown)


lib/statsd.js
```diff
@@ -402,7 +348,11 @@ Client.prototype.sendMessage = function (message, callback) {
...
   }
 
-  if (this.protocol === TCP && message.lastIndexOf('\n') !== message.length - 1) {
+  if (!this.socket) {
+    return callback(new Error('No transport set. Check previous errors for details.'));
+  }
+
+  if (this.protocol === PROTOCOL.TCP && message.lastIndexOf('\n') !== message.length - 1) {
     message += '\n';
   }
```
msiebuhr: I think this would be a better fit inside lib/transport.js

lib/statsd.js
```diff
@@ -527,19 +470,7 @@ Client.prototype._close = function (callback) {
...
-      // this close is synchronous, and the socket will not emit a
-      // close event, so we can callback right away
-      if (callback) {
-        callback();
-      }
-    }
-    else {
-      this.socket.close();
-    }
+    this.socket.close();
```
msiebuhr: How about also setting `this.socket = undefined`, so there's no doubt things are broken?

(And then it's trivial to do `if (!this.socket) { this.socket = createSocket(...) };` to attempt self-healing later on :)

.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "singleQuote": true
+}
```
msiebuhr: While I can't speak for the project, random dotfiles is a smell to me. AFAIK, this project doesn't use https://prettier.io/.

runk: Yeah, my bad - gonna kill it.

lib/statsd.js
runk: Good point!

lib/statsd.js
runk: Well, I'm trying not to change the logic in this PR and leaving behaviour as is. I better leave it intact.

Another place that bugs me a lot is https://github.com/brightcove/hot-shots/pull/129/files#diff-6b52e4167496e1de6769211f454200f7R80 - basically swallowing errors sometimes. Wish we'd just threw it, at least at setup stage.

lib/statsd.js
```diff
@@ -402,8 +348,8 @@ Client.prototype.sendMessage = function (message, callback) {
     return;
   }
 
-  if (this.protocol === TCP && message.lastIndexOf('\n') !== message.length - 1) {
```
runk: This has been moved to the tcp transport adapter

lib/statsd.js
```diff
@@ -160,7 +106,7 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   // will gracefully (attempt) to re-open the socket with a small delay
   // options.udsGracefulRestartRateLimit is the minimum time (ms) between creating sockets
   // does not support options.isChild (how to re-create a socket you didn't create?)
-  if (!options.isChild && options.protocol === 'uds' && options.udsGracefulErrorHandling) {
+  if (!options.isChild && options.protocol === PROTOCOL.UDS && options.udsGracefulErrorHandling) {
```
runk: Would be nice to abstract this piece in the uds transport, but perhaps a story for another PR. This is very special case for uds only.

bdeitte:
Thanks @runk!  @msiebuhr I can tell you that I very much welcome the idea of having more reviews of things and having other people who help maintain.  So always welcome to do this, thanks.  :)

lib/constants.js
```diff
@@ -0,0 +1,6 @@
+exports.PROTOCOL = {
+  TCP: 'tcp',
+  UDS: 'uds',
+  UDP: 'udp',
+  STREAM: 'stream'
```
bdeitte: Please add stream to the README

lib/statsd.js
```diff
@@ -402,8 +348,8 @@ Client.prototype.sendMessage = function (message, callback) {
     return;
   }
 
-  if (this.protocol === TCP && message.lastIndexOf('\n') !== message.length - 1) {
-    message += '\n';
+  if (!this.socket) {
+    return callback(new Error('No transport set. Check previous errors for details.'));
```
bdeitte:
```suggestion
    return callback(new Error('No protocol set. Check previous errors for details.'));
```

Since protocol is still the name exposed to users

lib/statsd.js
```diff
@@ -418,23 +364,16 @@ Client.prototype.sendMessage = function (message, callback) {
       if (this.errorHandler) {
         this.errorHandler(errFormatted);
       } else {
-        console.log(String(errFormatted));
+        console.error(String(errFormatted));
```
bdeitte: Thanks for the extra fix here

lib/constants.js
bdeitte: I understand now reading farther down that you haven't added this in yet as an actual option, that this would be a follow on PR or still coming

test/init.js
```diff
@@ -231,22 +228,4 @@ describe('#init', () => {
     statsd = createHotShotsClient(['host', 1234, 'prefix', 'suffix', false, false, true], clientType);
     assert.ok(statsd.mock);
   });
-
-  it('should create a socket variable that is an instance of dgram.Socket', () => {
-    statsd = createHotShotsClient({}, clientType);
-    assert.ok(statsd.socket instanceof dgram.Socket);
-    skipClose = true;
-  });
```
bdeitte: Is there a reason this was removed?  In the past people did access this directly, for better or worse, to catch errors, so this would become a backwards-compatibility issue to lose.

lib/constants.js
runk: yes, that's the plan

bdeitte:
A couple comments.  I really like the direction of abstracting this out.

test/init.js
runk: Well, I was unsure what to do since `socket` is not an instance of the net/dgram anymore.

runk:
@bdeitte @msiebuhr can you please check it out again?

bdeitte:
Thanks, looks good to me @runk.  I'll wait to see if @msiebuhr has any more comments and then can merge in.

test/helpers/helpers.js
```diff
@@ -5,7 +5,7 @@ const fs = require('fs');
 const os = require('os');
 const process = require('process');
 const StatsD = require('../../lib/statsd.js');
-
+const { PROTOCOL: { TCP, UDP, UDS } } = require('../../lib/constants');
```
msiebuhr: I'd personally prefer `const { TCP, UDP, UDS } = require('../../lib/constants').PROTOCOL;` - I find multi-layered unwrapping needlessly hard to follow. KISS and all...

runk:
@bdeitte should be ready to go

PR #130, opened 2019-10-10
Stream transport

runk:


runk:
Sorry for the amount of changes, but cannot get away without touching every single test file. 😢

lib/statsd.js
```diff
@@ -64,7 +65,7 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   this.telegraf = options.telegraf || false;
   this.maxBufferSize = options.maxBufferSize || 0;
   this.sampleRate = options.sampleRate || 1;
-  this.bufferFlushInterval = options.bufferFlushInterval || 1000;
+  this.bufferFlushInterval = options.bufferFlushInterval || (process.env.NODE_ENV === 'test' ? 10 : 1000);
```
runk: This makes tests to run a lot faster.

package.json
```diff
@@ -30,8 +30,8 @@
   },
   "scripts": {
     "coverage": "nyc --reporter=lcov --reporter=text npm test",
-    "test": "mocha --exit -R spec --timeout 5000 test/*.js",
-    "lint": "eslint lib/**.js test/**.js",
+    "test": "NODE_ENV=test mocha --exit -R spec --timeout 5000 test/*.js",
+    "lint": "eslint lib/**/*.js test/**/*.js",
```
runk: This makes linter to run for all files in all dirs

runk:
@msiebuhr @bdeitte any feedback on the above?

lib/statsd.js
msiebuhr: Then I think the tests should set it explicitly -- I hate it (as application developer) when underlying libraries changes behaviour between environments...

package.json
msiebuhr: I thought `npm test` implied setting `NODE_ENV=test`...

runk: Sorry for misleading comment, I had to go different path: https://github.com/brightcove/hot-shots/pull/130/files#diff-831d41ad895e89c8dffa55c52a8bd1bcR112

lib/statsd.js
runk: As mentioned - now it's set explicitly in testing/mocking code.

runk:
@bdeitte any thoughts?

bdeitte:
@runk Looks good to me.  I can't merge right now, but I'll do so later today unless I hear any more feedback before then from @msiebuhr  or others.

PR #133, opened 2019-10-16
Add unref method to transport interface

### LINKED ISSUE #132: New transport api makes it impossible to unref sockets
related to #57 but the workarounds from that issue were broken the most recent set of changes

hot-shots keeps an open socket/connection that will never close under normal use.   This prevents any node process with an open client from exiting cleanly. 

Previously we would just do something like `client.socket.unref()` or unref the socket in a subclass.  Unfortunately the new abstracted transport API does not allow unrefing sockets since they are hidden in a closure.

hayes:
fixes #132

### DIFF
--- a/lib/transport.js
+++ b/lib/transport.js
@@ -22,6 +22,7 @@ const addEol = (buf) => {
 //   removeListener(name: string, listener: Function):void;
 //   send(buf: Buffer, callback: Function):void;
 //   close():void;
+//   unref(): void;
 // }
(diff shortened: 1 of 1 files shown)


bdeitte:
This looks reasonable to me.  I would ask for a test, but since we weren't testing but supporting already, I'll ignore that for now.  :)  Certainly want to make sure people aren't broken here, so merging in.

hayes:
Thanks!

Issue #136, opened 2019-11-12
useDefaultRoute is not working from v6.7.0


almandsky:
https://github.com/brightcove/hot-shots/compare/v6.6.0...v6.7.0

Seems like the `createTransport` is taking the `host` parameter before it is updated by the `useDefaultRoute`.  That means, from v6.7.0, if the hot-shots is running inside a docker container, it will still send the metrics to localhost by default, instead of sending to the actual host ip `172.17.0.1`.

We shall consider move the code of line 146~152 to before line 43, which is where the `host` is being used for `createTransport`.

https://github.com/brightcove/hot-shots/blob/master/lib/statsd.js#L146-L152

Gobie:
`cacheDns` option doesn't work neither because of the same problem.

almandsky:
Before the v6.7.0, the `Client.prototype.sendMessage` will use the `host` args after the init is completed.

But from v6.7.0, the `host` is passed to the `createTransport` during the init, so the `host` override will not work after the `this.socket` is initialized via `createTransport`.

https://github.com/brightcove/hot-shots/compare/v6.5.1...v6.8.1#diff-cfdf1e0afcbe7fb32697031bec3cc7deL393

almandsky:
@Gobie 

I include the fix for `cacheDns` in my PR as well.

Gobie:
Hey @almandsky, thanks for taking the time to tackle this issue.
But I believe `cacheDns` still remains broken as `dns.lookup` is an async call and by the time it is resolved `createTransport` have already been called and `host` can't be overridden.

almandsky:
@Gobie 

I don't think `dns.lookup` is a async call.  It is still a sync call, but you can provide the callback function to it.

https://git.softwareadvice.com/qa/testcafe/blob/db29579085112bfa27808bbe680e2d3747c42ad8/node_modules/@types/node/dns.d.ts#L30

Gobie:
I would respectfully disagree
```js
const dns = require('dns')
dns.lookup('google.com', console.log.bind(console, 'lookup callback'))
console.log('call after lookup')
```

```bash
➜ node jest/test.js
call after lookup
lookup callback null 216.58.201.110 4
```

almandsky:
@Gobie 

Sorry that I am wrong.  I read some posts before saying that the dns.lookup was blocking the main process.  Thanks for the correction.

[...]

PR #137, opened 2019-11-12
override host before socket created by createTransport

almandsky:
As per https://github.com/brightcove/hot-shots/issues/136

The `useDefaultRoute` was being handled after the `host` is being used for `createTransport`.

This PR is to move the `useDefaultRoute` above to override the `this.host` before it is being used. Also move the `this.socket` init after all the `host` override so that the `cacheDns` will work as well.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -41,6 +41,14 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   // hidden global_tags option for backwards compatibility
   options.globalTags = options.globalTags || options.global_tags;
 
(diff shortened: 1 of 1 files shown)


bdeitte:
Thanks for the fix.  Bummer, looks like we don't have any existing test for this, so that's how it was missed, and not sure how to add one exactly quickly.  I don't want to hold up a fix for that certainly, and looks good to me.

almandsky:
@bdeitte 

Thanks for the review and merged the PR.  I tested the v6.8.2 that run inside docker container, and it works fine with the `useDefaultRoute: true` option.

Issue #140, opened 2019-12-10
TypeScript StatsD type issues: Type 'StatsD' has no construct signatures.


gleb-rudenko:
Using typescript I import StatsD:
`import StatsD from "hot-shots";`

When I want to create a new client instance like this:
`const client = new StatsD({...});`
I get `ts(2351)` error, that says:
```
This expression is not constructable.
  Type 'StatsD' has no construct signatures.
```

I don't seem to get what's the problem here, is that a known issue? `types.d.ts` looks fine, if I didn't miss something.

bdeitte:
Thanks for the issue and fix here @gleb-rudenko.  I wonder why others who have been using the TypeScript setup haven't run into this, or if it's just lost in TypeScript error noise.  To ask a few who have submitted changes there, @drewen @stieg @etaoins, just want to make sure this makes sense to you too.

etaoins:
I've been using this which compiles fine:

```typescript
import { StatsD } from "hot-shots"
```

The patch looks correct to me, however.

bdeitte:
Fixed in 6.8.2

gleb-rudenko:
Hi @etaoins ,
all of the docs are targeted to JS so every example does default import whether it common JS or es6 imports. It looks inconvenient to use different import behavior for TS, that's why I was confused with that.

Btw, thanks for merging my PR, @bdeitte :)

PR #141, opened 2019-12-10
Fix StatsD typings

gleb-rudenko:
As per #140 changed the `export class StatsD`  to be an interface.


bdeitte:
I added a comment and checking on something in https://github.com/brightcove/hot-shots/issues/140 and then I can merge in.

bdeitte:
And since it had a review now, merging in, thanks

Issue #138, opened 2019-11-12
cacheDns is not working from v6.7.0 


bdeitte:
See details at https://github.com/brightcove/hot-shots/issues/136#issuecomment-553016683

bdeitte:
Fixed in 6.8.4

Issue #63, opened 2018-04-07
Add mocking in tests


bdeitte:
Use sinon or something similar to test the parts that are hard to test right now, and to actually unit test.  One good candidate to start testing here would be getDefaultRoute()

bdeitte:
With this is place we could also test the sending of calls properly with cacheDns

bdeitte:
Fixed in 13.0.0

PR #143, opened 2019-12-15
Fix cacheDns with udp

bdeitte:
Fix https://github.com/brightcove/hot-shots/issues/138

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -16,7 +16,6 @@ const UDS_DEFAULT_GRACEFUL_RESTART_LIMIT = 1000;
 const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
     globalTags, maxBufferSize, bufferFlushInterval, telegraf, sampleRate, protocol) {
   let options = host || {};
-  const self = this;
 
   // Adding options below is DEPRECATED.  Use the options object instead.
   if (arguments.length > 1 || typeof(host) === 'string') {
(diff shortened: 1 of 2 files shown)


lib/statsd.js
```diff
@@ -16,7 +16,6 @@ const UDS_DEFAULT_GRACEFUL_RESTART_LIMIT = 1000;
 const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
     globalTags, maxBufferSize, bufferFlushInterval, telegraf, sampleRate, protocol) {
   let options = host || {};
-  const self = this;
```
bdeitte: Don't see this needed here so simplified below

lib/statsd.js
```diff
@@ -77,32 +76,39 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
     if (options.dnsError) {
       this.dnsError = options.dnsError;
     }
+    this.socket = options.socket;
```
bdeitte: This is taken from what used to be: 
`this.socket = options.isChild ? options.socket : createTransport(this, {`
The creation of the socket is moved up into this if block so that the transport is created at the right time.  This changes the creation of the transport for the cacheDns case to happen after the dns.lookup is done below so that it has the correct host.

test/init.js
```diff
@@ -147,7 +147,7 @@ describe('#init', () => {
     assert.deepEqual(statsd.globalTags, ['gtag', 'dd.internal.entity_id:04652bb7-19b7-11e9-9cc6-42010a9c016d']);
   });
 
-  it('should attempt to cache a dns record if dnsCache is specified', done => {
+  it('should lookup a dns record if dnsCache is specified', done => {
```
bdeitte: Make clear this isn't really testing the attempt to cache, just the lookup.  I tried to add a test for this bug, but I didn't see a straightforward way without mocking in place.  Making a note in https://github.com/brightcove/hot-shots/issues/63

bdeitte:
Ok let this sit for a few days and merging in now

Issue #150, opened 2020-01-28
callback is not properly passed bytes arguement 


ericmustin:
Hey there,

Hope all is well. The use of a callback, which is supported in your documentation and well as in the underlying `dgram` package, appears to be improperly implemented starting in Jan'19 with [this commit]( https://github.com/brightcove/hot-shots/commit/c49197c2adb451bbd8e62641c1918c1b4ee97009)

in the example
```
  // with all non-close calls
  client.set(['foo', 'bar'], 42, function(error, bytes){
    //this only gets called once after all messages have been sent
    if(error){
      console.error('Oh noes! There was an error:', error);
    } else {
      console.log('Successfully sent', bytes, 'bytes');
    }
  });
```
the `bytes` argument will always be undefined.

You can reproduce this with the below snippet:

```
var StatsD = require('hot-shots')
var client = new StatsD({
      port: 8125,
      globalTags: { env: process.env.NODE_ENV },
      errorHandler: function(error){console.log('ok', error)},
      protocol: 'udp'
  });
 // Using the callback.  This is the same format for the callback
// with all non-close calls
client.set(['foo', 'bar'], 42, function(error, bytes){
  //this only gets called once after all messages have been sent
  if(error){
    console.error('Oh noes! There was an error:', error);
  } else {
    console.log('Successfully sent', bytes, 'bytes');
  }
});
```
=> `Successfully sent 0 bytes`

Specifically, it looks like this helper method [`handleCallback`](https://github.com/brightcove/hot-shots/blob/daccafd248e29f3b23db83e3dbb2251def2e20b7/lib/statsd.js#L365) needs to have a second parameter defined, the `bytes` param that is mentioned in your docs. By updating to the below, the issue can be resolved.

```
  const handleCallback = (err, bytes) => {
    this.messagesInFlight--;
    const errFormatted = err ? new Error(`Error sending hot-shots message: ${err}`) : null;
    if (errFormatted) {
      errFormatted.code = err.code;
    }
    if (callback) {
      callback(errFormatted, bytes);
    } else if (errFormatted) {
      if (this.errorHandler) {
        this.errorHandler(errFormatted);
      } else {
        console.error(String(errFormatted));
        // emit error ourselves on the socket for backwards compatibility
        this.socket.emit('error', errFormatted);
      }
    }
  };
```

I can make a PR as soon as I get a free moment here

bdeitte:
[...]

PR #151, opened 2020-01-28
pass in bytes as arg to handleCallback

ericmustin:
Fixes https://github.com/brightcove/hot-shots/issues/150

Please let me know if there's any more information needed here, the bug, reproduction, and discussion can be found on attached open issue.

Happy to implement any feedback as you see fit! Cheers


ericmustin:
Much appreciated @bdeitte , I agree with what you’re saying. I can get some tests added here as soon as I have bandwidth. Thank you on the very very quick turn around

bdeitte:
And thanks for the fix and tests coming, appreciate it!

PR #153, opened 2020-02-05
Check callback is set when socket not set

### LINKED ISSUE #152: TypeError: callback is not a function
Hi,

We're running into an issue when we call `increment` without a callback argument, e.g.:

```
baseStatsD.increment('exception-handler.exception', 1, tags)
```

and the socket is `null` (this can happen if the socket is not successfully created by the metric server - in our case Datadog agent), `callback` is called with an error (https://github.com/brightcove/hot-shots/blob/master/lib/statsd.js#L361-L363).

However, callback is undefined so this fails with the following trace:

Full trace:
```
TypeError: callback is not a function 
at Client.sendMessage (/usr/src/app/node_modules/hot-shots/lib/statsd.js:362:12)
at Client._send (/usr/src/app/node_modules/hot-shots/lib/statsd.js:306:12)
at Client.send (/usr/src/app/node_modules/hot-shots/lib/statsd.js:283:8)
at Client.sendStat (/usr/src/app/node_modules/hot-shots/lib/statsd.js:260:8)
at Client.sendAll (/usr/src/app/node_modules/hot-shots/lib/statsd.js:236:10)
at Client.increment (/usr/src/app/node_modules/hot-shots/lib/statsFunctions.js:114:10) 
```

mrknmc:
Fixes #152 

We check if callback is set before calling it with socket error and if it is not we log an error instead.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -358,8 +358,14 @@ Client.prototype.sendMessage = function (message, callback) {
     return;
   }
 
-  if (!this.socket && callback) {
-    return callback(new Error('Socket not created properly. Check previous errors for details.'));
+  if (!this.socket) {
+    var error = new Error('Socket not created properly. Check previous errors for details.');
+    if (callback) {
+      return callback(error);
+    } else {
+        console.error(String(error));
+        return;
+    }
   }
 
   const handleCallback = (err, bytes) => {
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -359,7 +359,7 @@ Client.prototype.sendMessage = function (message, callback) {
   }
 
   if (!this.socket) {
-    var error = new Error('Socket not created properly. Check previous errors for details.');
+    let error = new Error('Socket not created properly. Check previous errors for details.');
     if (callback) {
       return callback(error);
     } else {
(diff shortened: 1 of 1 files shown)


lib/statsd.js
```diff
@@ -359,7 +359,13 @@ Client.prototype.sendMessage = function (message, callback) {
   }
 
   if (!this.socket) {
-    return callback(new Error('Socket not created properly. Check previous errors for details.'));
+    const error = 'Socket not created properly. Check previous errors for details.';
+    if (callback) {
+      return callback(new Error(error));
+    } else {
+        console.error(error);
+        return;
```
bdeitte: Can you remove this "return", as the style in here hasn't been to include in this case, and can you unindent the console.error line by 2?

Thanks!  Once that is done I can merge this in, very clear explanation you gave in the bug on what is wrong here.

mrknmc: Hey @bdeitte 

I fixed the indent and removed the return in bd2b68e5ae5f2c84f2fa2748371b58136dafcb46.

However, without the return, this is going to error out on the [socket send](https://github.com/brightcove/hot-shots/pull/153/files#diff-cfdf1e0afcbe7fb32697031bec3cc7deR391). Is that what we want?

bdeitte: Oh you're right there, the inline function got me there!  I'll add back in as I merge, thanks.

PR #154, opened 2020-02-28
Parameterize function types in timer and asyncTimer

casey-chow:
This allows Typescript to transparently typecheck the wrapped function for `timer` and `asyncTimer` rather than downcasting from the `any` type, allowing greater type safety.


bdeitte:
Thanks, this looks fine to me, merging in

PR #155, opened 2020-03-04
Fix missing host update

wision:
Recent fixes for DNS cache did not help because the host has not been properly updated. This fix was tested with tcpdump.


bdeitte:
Shoot, yes that was my mistake.  Thanks for the fix!

wision:
Thank you. I also suggest to fix the test.

PR #156, opened 2020-03-06
Don't send undefined as a value

wision:
I'm not really sure that this should be part of this repository, so I'll leave that up to you. The problem is that when you send `undefined` as a value, statsd will not store anything and hence it received redundant message which had to be processed and it will generate error in the log:
```
2020-03-06T09:56:29.199975+00:00 statsd1 statsd[30923]: Bad line: undefined,g in msg "servers.app03.application.4d2a665c-5ed0-11ea-826a-06cd01d45cce.io_read:undefined|g"
```

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -197,6 +197,13 @@ Client.prototype.sendAll = function (stat, value, type, sampleRate, tags, callba
     tags = undefined;
   }
 
+  if (value === undefined) {
+    if (typeof callback === 'function') {
+      callback(null, 0);
+    }
+    return;
+  }
+
   /**
    * Gets called once for each callback, when all callbacks return we will
    * call back from the function


lib/statsd.js
```diff
@@ -197,6 +197,13 @@ Client.prototype.sendAll = function (stat, value, type, sampleRate, tags, callba
     tags = undefined;
   }
 
+  if (value === undefined) {
+    if (typeof callback === 'function') {
+      callback(null, 0);
+    }
+    return;
+  }
```
bdeitte: This feels like an error case, since nothing is being sent as expected.  I would be ok with this if it sends on an error message in the callback and called errorHandler if that was present, similar to the error code below.

I also would like to see a test for this.

If those things can be changed, I'm good with adding this.  I would want to let it sit in review for a little bit, just in case someone else knows of a case (with DogStatsD or Telegraf) where this would want to send undefined, although I don't know of any myself.

wision: I wanted it to behave the same as before (it just wouldn't send anything at all), but I agree that we should see what happens in the other services.

bdeitte: I'm concerned however that it does not actually behave the same as before- without knowing how all endpoints react to this, we could be changing behavior.  It's also not sending when the function says it would, so I do want to see this go through the error case to merge this in.  Also I will need a test for the case.  Thanks.

wision: That's ok. We'll deal with this in our wrapper package. Thank you for your time.

Issue #104, opened 2019-03-27
Refresh cached dns periodically


DennisPan:
Currently, enabling `cacheDns` caches the address indefinitely. If the ip changes, then the messages are dropped. Fortunately, in UDP's case, `socket.send` does not throw if the ip is no longer there.

Wondering if this has been brought up before? If we implement something that does a new `dns.lookup` (e.g. every 30 seconds, or make it configurable), would that be a good solution? Happy to submit a PR for that.

bdeitte:
I haven't heard this brought up before, but it definitely makes sense that this could be an issue.  I think this should be configurable, and I think a higher-but-not-indefinite lookup would be good, perhaps 5 minutes?  Thanks for any PR on it!

msiebuhr:
I have a sort-of-similar problem where data sent to a Datadog-agent doesn't arrive after is has been re-scheduled (... inside containers ... inside Kubernetes; full story on [stack overflow](https://stackoverflow.com/questions/58094024/intra-host-udp-traffic-missing-after-destination-pod-re-created).

I considered doing a hotfix to force re-creation of the underlying socket every 30s or so:

```javascript
const statsd = new hotShots.Client(...);
setInterval(() => statsd.close(), 30 * 1000);
```

msiebuhr:
> I have a sort-of-similar problem where data sent to a Datadog-agent doesn't arrive after is has been re-scheduled (... inside containers ... inside Kubernetes; full story on [stack overflow](https://stackoverflow.com/questions/58094024/intra-host-udp-traffic-missing-after-destination-pod-re-created).
> 
> I considered doing a hotfix to force re-creation of the underlying socket every 30s or so:
> 
> ```js
> const statsd = new hotShots.Client(...);
> setInterval(() => statsd.close(), 30 * 1000);
> ```

Tried it out -- but as there's no way to re-connect (besides creating a whole new client), this doesn't work out terribly well.

marciopd:
Hello guys, I'm having this issue recurrently in a Kubernetes cluster.
I've created a [pull request](https://github.com/brightcove/hot-shots/pull/158) as a proposal to fix it.

The dns ttl is set as `60s`per default and can be configured through option `dnsTtlMilliseconds `.
If we want to keep backwards compatibility, of course we could change this.

I would appreciate some comments/suggestions.

bdeitte:
Released as option cacheDnsTtl in v7.2.0

PR #158, opened 2020-03-14
Enables a configurable time-to-live for dns lookups for UDP transport.

marciopd:
Hello,

Currently, when `cacheDns` is enabled, you'll have a single lookup when the client is created. The risk (which depending on your environment will happen sooner or later) is losing metrics when the target host changes the IP for whatever reason.

There is already an issue for this: https://github.com/brightcove/hot-shots/issues/104

This PR enables a TTL for dns lookups with `60s` as the default. 
Please consider if the changes are ok and let me know if you would like any changes.

Best regards!

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,9 @@ Parameters (specified as one object passed into hot-shots):
 * `prefix`:      What to prefix each stat name with `default: ''`
 * `suffix`:      What to suffix each stat name with `default: ''`
 * `globalize`:   Expose this StatsD instance globally. `default: false`
-* `cacheDns`:    Cache the initial dns lookup to *host* , only used
-  when protocol is `uds`, `default: false`
+* `cacheDns`:    Caches dns lookup to *host* for *dnsTtlMilliseconds*, only used
+  when protocol is `udp`, `default: false`
+* `dnsTtlMilliseconds`: time-to-live of dns lookups in milliseconds, when *cacheDns* is enabled. `default: 60000`
 * `mock`:        Create a mock StatsD instance, sending no stats to
   the server and allowing data to be read from mockBuffer.  Note that
   mockBuffer will keep growing, so only use for testing or clear out periodically. `default: false`
--- a/README.md
+++ b/README.md
@@ -35,9 +35,9 @@ Parameters (specified as one object passed into hot-shots):
 * `prefix`:      What to prefix each stat name with `default: ''`
 * `suffix`:      What to suffix each stat name with `default: ''`
 * `globalize`:   Expose this StatsD instance globally. `default: false`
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -35,8 +35,9 @@ Parameters (specified as one object passed into hot-shots):
 * `prefix`:      What to prefix each stat name with `default: ''`
 * `suffix`:      What to suffix each stat name with `default: ''`
 * `globalize`:   Expose this StatsD instance globally. `default: false`
-* `cacheDns`:    Cache the initial dns lookup to *host* , only used
-  when protocol is `uds`, `default: false`
+* `cacheDns`:    Caches dns lookup to *host* for *dnsTtlMilliseconds*, only used
+  when protocol is `udp`, `default: false`
+* `dnsTtlMilliseconds`: time-to-live of dns lookups in milliseconds, when *cacheDns* is enabled. `default: 60000`
```
bdeitte: We don't include the time type in other names, so could this name be changed to be dnsTtl?  Actually, to aid even more in understanding it's a part of cacheDns, how about cacheDnsTtl?

lib/statsd.js
```diff
@@ -44,6 +43,8 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   if (! this.protocol) {
     this.protocol = PROTOCOL.UDP;
   }
+  this.cacheDns = options.cacheDns === true;
+  this.dnsTtlMilliseconds = options.dnsTtlMilliseconds || 60000;
```
bdeitte: Can you add 60000 as a const above?  I know, I know, very much not in place in here.  :) Trying to have new code start to do this however.

test/udpDnsCacheTransport.js
```diff
@@ -0,0 +1,158 @@
...
+              assert.equal(dnsLookupCount, 2);
+              assert.equal(socketMock.sendCount, 3);
+              done();
+            }, 1000);
+          });
+        }, dnsTtlMilliseconds + 1);
+      });
+    });
+  });
+});
```
bdeitte: Great to see, thanks

bdeitte:
Thanks for the PR!  Two requests and then I'm good to merge.

README.md
marciopd: Sure, changed it to `cacheDnsTtl`.

lib/statsd.js
marciopd: Sure, I understand 😄 . Done, please check.

test/udpDnsCacheTransport.js
marciopd: You're welcome! Thanks for looking into this!

bdeitte:
Looks good, merging it in and creating 7.2.0

PR #161, opened 2020-03-31
Tag prefix separator

MichaelSitter:
Hello! Big fan of the library but we've been struggling to use it to it's fullest because of some idiosyncrasies of our statsd format, particularly with tags. In our case it can be addressed by allowing control over the characters used to separate tags:

```
* `tagPrefix`:   Prefix tag list with character `default: '#'`. Note does not work with `telegraf` option.
* `tagSeparator`: Separate tags with character `default: ','`. Note does not work with `telegraf` option.
```

I implemented this in a fork for my own use, but wanted to see if there was any interest in using it. Cheers!


bdeitte:
Glad to hear it!  This looks perfectly reasonable to me to include.  There is a conflict now with other changes merged- can you fix that?  Thanks.

MichaelSitter:
Thanks @bdeitte! Conflict resolved.

bdeitte:
Looks good, I'll release in 7.4.0

PR #166, opened 2020-05-04
Fix types for set/unique

kazk:
Value can be a string.

```javascript
// Set: Counts unique occurrences of a stat (alias of unique)
client.set('my_unique', 'foobar');
client.unique('my_unique', 'foobarbaz');
```


bdeitte:
Makes sense, thanks

PR #167, opened 2020-06-01
Fix socket/interval blocking

benblack86:
Need to call `unref` on sockets and interval timers so node isn't blocked from shutting down. I noticed this problem as my tests weren't exiting. Since this fixes the problem I removed `--exit` from mocha options as we don't need to force exit anymore.

https://nodejs.org/api/timers.html#timers_timeout_unref
https://nodejs.org/api/net.html#net_socket_unref

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -227,7 +227,6 @@ module.exports = {
         "prefer-const": "error",
         "prefer-destructuring": "off",
         "prefer-numeric-literals": "error",
-        "prefer-object-spread": "error",
         "prefer-promise-reject-errors": "error",
         "prefer-reflect": "off",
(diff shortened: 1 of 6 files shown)


.eslintrc.js
```diff
@@ -227,7 +227,6 @@ module.exports = {
         "prefer-const": "error",
         "prefer-destructuring": "off",
         "prefer-numeric-literals": "error",
-        "prefer-object-spread": "error",
```
benblack86, line 230: It wouldn't let me do `{ ...statsd.socket }` so I removed the rule.

bdeitte:
This looks great, thanks for doing this.  I'm going to put it in a minor release shortly.

PR #169, opened 2020-06-10
add date type to timing function

### LINKED ISSUE #168: client.timing could accept date
currently, the timing method just accepts timestamp which is time in milliseconds, which could be more convenient if a date is also accepted.

Will raise a PR

Impeekay:
tests needed.

fixes #168

### DIFF
--- a/test/statsFunctions.js
+++ b/test/statsFunctions.js
@@ -143,7 +143,7 @@ describe('#statsFunctions', () => {
             done();
           });
         });
-      })
+      });
 
       describe('#increment', () => {
         it('should send count by 1 when no params are specified', done => {
--- a/test/statsFunctions.js
+++ b/test/statsFunctions.js
@@ -17,13 +17,13 @@ describe('#statsFunctions', () => {
   testTypes().forEach(([description, serverType, clientType, metricsEnd]) => {
     describe(description, () => {
       [{ name: 'timing', unit: 'ms', bytes: 14 },
-       { name: 'histogram', unit: 'h', bytes: 12 },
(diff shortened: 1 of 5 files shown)


bdeitte:
Thanks for the addition.  I would like to see a test added for this before bringing in, so that we don't break it later.  You could add a very simple one around here: https://github.com/brightcove/hot-shots/blob/master/test/statsFunctions.js#L125

And two other requests here.  Could you add "(or Date object of which the difference is calculated)" to the comment line in READMEm.md, to indicate this new functionality.  And change "number" to "number | Date" for the four type lines for this function: https://github.com/brightcove/hot-shots/blob/master/types.d.ts#L88

Thanks!

Impeekay:
Sure, will update the PR.

test/statsFunctions.js
```diff
@@ -122,6 +122,29 @@ describe('#statsFunctions', () => {
...
+          server = createServer(serverType, opts => {
+            statsd = createHotShotsClient(opts, clientType);
+            statsd.timing('test', new Date(Date.now() - 10));
+          });
+          server.on('metrics', metrics => {
+            assert.equal(metrics, `test:10|ms${metricsEnd}`);
+            done();
+          });
+        });
+      })
```
bdeitte, line 146: This is missing a semi-colon (which is causing the tests to fail).  Once this is fixed, I can merge this all in, thanks!

Impeekay: Damn, did not see that. It did pass on mine when I tested it. Also it'd be better if we put out a prettier config for ppl to use. :) 
I'll update it.

PR #171, opened 2020-06-25
Add asyncDistTimer function

tebriel:
We're heavily moving into Datadog distributions. `asyncTimer` reports a `timing` metric (type `ms`) which is eventually a histogram in datadog. To avoid confusion (and to allow us to emit a distribution with the asyncTimer) I've created a new function called asyncDistTimer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ includes all changes in the latest node-statsd and many additional changes, incl
 * raw stream protocol support
 * mock mode
 * asyncTimer
+* asyncDistTimer
 * much more, including many bug fixes
 
 hot-shots supports Node 8.x and higher.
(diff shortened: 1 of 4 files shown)


tebriel:
Looks like the check is failing due to a timing issue, a re-run will probably fix it!

README.md
```diff
@@ -170,14 +171,21 @@ The check method has the following API:
   var fn = function(a, b) { return a + b };
   client.timer(fn, 'fn_execution_time')(2, 2);
 
-  // Async timer: Similar to timer above, but you instead pass in a funtion
+  // Async timer: Similar to timer above, but you instead pass in a function
```
bdeitte, line 174: Thanks for the misc fixup, and the example below, looks good

bdeitte:
Thanks for the addition here.  I agree it was just a timing issue for the test, hopefully those don't start creeping up again (thought I took care of them generally).  All looks good, will create a minor release shortly.

Issue #172, opened 2020-07-02
udsGracefulErrorHandling default isn't used


DerGut:
The default value of isn't actually used and error handling only happens when `udsGracefulErrorHandling` is explicitly set to  `true` (https://github.com/brightcove/hot-shots/issues/131#issuecomment-652574618).

PR #173, opened 2020-07-02
Fix udsGracefulErrorHandling default value

DerGut:
Fixes https://github.com/brightcove/hot-shots/issues/172 where `udsGracefulErrorHandling` is only used when explicitly set in the options.


bdeitte:
Looks great, thanks!

PR #177, opened 2020-07-28
Add option to snake case tags before sending

casey-chow:
One thing I found while using this package is that my codebase enforces camelCase naming of object properties, but datadog lowercases everything so snake_case is generally ideal. This adds an option to automatically snake case tags before sending out.


bdeitte:
Hi @casey-chow thanks for the PR.  I'm usually a bit faster at reviewing here but had some time off the laptop here and just catching up today. 

I really try not to add new dependencies in here, given the wide usage of this library and my own wishes to keep the package list small.  For this use case, I would need to see the underscore functionality as a function in here to accept this.  There would also need to be some tests added.  Thanks.

bdeitte:
Hi @casey-chow just checking on this, whether it's still on your radar or something to close out.  Thanks.

bdeitte:
@casey-chow closing out, but please feel free to reopen if changes are made, thanks.

PR #186, opened 2020-09-09
replace 'localhost' fallback with undefined

### LINKED ISSUE #185: feat: use 127.0.0.1 instead of localhost
https://github.com/brightcove/hot-shots/blob/5ec98a55c05e0c490ccf993a0a768f5364683fc6/lib/statsd.js#L49

From https://nodejs.org/api/dgram.html#dgram_socket_send_msg_offset_length_port_address_callback

>  The address argument is a string. If the value of address is a host name, DNS will be used to resolve the address of the host. If address is not provided or otherwise falsy, '127.0.0.1' (for udp4 sockets) or '::1' (for udp6 sockets) will be used by default.

By changing this we avoid calling DNS to resolve `localhost`

naseemkullah:
fixes #185 

According to https://nodejs.org/api/dgram.html#dgram_socket_send_msg_offset_length_port_address_callback, if the address provided is a hostname such as localhost, it will query DNS to resolve it. Setting 127.0.0.1 will avoid this DNS lookup when running the agent on localhost

**edit**: instead of hardcoding 127.0.0.1 we will go with leaving undefined.

According to dgram socket, if the address provided is a hostname such as
localhost, it will query DNS to resolve it. dns.lookup() has a fast path
for IP addresses, when omitting a value for host.

Also, from
https://nodejs.org/api/dgram.html#dgram_socket_send_msg_offset_length_port_address_callback:

"If address is not provided or otherwise falsy, '127.0.0.1' (for udp4
sockets) or '::1' (for udp6 sockets) will be used by default."

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,9 @@ You can check the detailed [change log](https://github.com/brightcove/hot-shots/
 All initialization parameters are optional.
 
 Parameters (specified as one object passed into hot-shots):
-* `host`:        The host to send stats to, if not set, the constructor tries to retrieve it from the `DD_AGENT_HOST` environment variable, `default: localhost`
+
+* `host`:        The host to send stats to, if not set, the constructor tries to
+  retrieve it from the `DD_AGENT_HOST` environment variable, `default: 'undefined'` which as per [UDP/datagram socket docs](https://nodejs.org/api/dgram.html#dgram_socket_send_msg_offset_length_port_address_callback) results in `127.0.0.1` or `::1` being used.
 * `port`:        The port to send stats to, if not set, the constructor tries to retrieve it from the `DD_DOGSTATSD_PORT` environment variable, `default: 8125`
 * `prefix`:      What to prefix each stat name with `default: ''`
 * `suffix`:      What to suffix each stat name with `default: ''`
--- a/test/init.js
+++ b/test/init.js
@@ -37,18 +37,18 @@ describe('#init', () => {
       clientType
     );
 
-    assert.equal(statsd.host, 'host');
-    assert.equal(statsd.port, 1234);
-    assert.equal(statsd.prefix, 'prefix');
-    assert.equal(statsd.suffix, 'suffix');
-    assert.equal(statsd, global.statsd);
-    assert.equal(statsd.mock, true);
-    assert.deepEqual(statsd.globalTags, ['gtag']);
-    assert.equal(statsd.maxBufferSize, 0);
-    assert.equal(statsd.bufferFlushInterval, 60);
-    assert.equal(statsd.telegraf, false);
-    assert.equal(statsd.sampleRate, 0.5);
-    assert.equal(statsd.protocol, 'udp');
+    assert.strictEqual(statsd.host, 'host');
+    assert.strictEqual(statsd.port, 1234);
+    assert.strictEqual(statsd.prefix, 'prefix');
+    assert.strictEqual(statsd.suffix, 'suffix');
+    assert.strictEqual(statsd, global.statsd);
+    assert.strictEqual(statsd.mock, true);
+    assert.deepStrictEqual(statsd.globalTags, ['gtag']);
+    assert.strictEqual(statsd.maxBufferSize, 0);
+    assert.strictEqual(statsd.bufferFlushInterval, 60);
+    assert.strictEqual(statsd.telegraf, false);
+    assert.strictEqual(statsd.sampleRate, 0.5);
+    assert.strictEqual(statsd.protocol, 'udp');
   });
 
   it('should set the proper values with options hash format', () => {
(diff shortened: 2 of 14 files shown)


naseemkullah:
edit: `dns.lookup()` will not be avoided completely, but please see https://github.com/nodejs/node/issues/35130#issuecomment-690082332

> dns.lookup() has a fast path for IP addresses

bdeitte:
Hi @naseemkullah thanks for the PR!  I'm a little hesitant to merge this in for a few reasons:
1. It's not clear to me what happens for udp6 sockets, if this will break
2. It's a really base setting, and without seeing performance metrics I'm worried that it might not behave as intended, that there's some other gotcha in here

On #2, have you tried this with DD_AGENT_HOST and seen a difference?  Thanks.

naseemkullah:
> Hi @naseemkullah thanks for the PR! I'm a little hesitant to merge this in for a few reasons:
> 
> 1. It's not clear to me what happens for udp6 sockets, if this will break
> 2. It's a really base setting, and without seeing performance metrics I'm worried that it might not behave as intended, that there's some other gotcha in here
> 
> On #2, have you tried this with DD_AGENT_HOST and seen a difference? Thanks.

Thanks @bdeitte I understand your concern about udp6. I've rarely used ipv6 so cannot comment much to that matter.

FWIW re: #2, setting DD_AGENT_HOST to 127.0.0.1 did result in significant performance improvements for our kubernetes service in question.

bdeitte:
Well that's really good to hear about #2.  I will assume then in general #2 is better, or at least I could bump this in a major version and feel good about releasing.  But that's only if someone can figure out what happens with #1.  Not sure what happens there still.

bdeitte:
So I see now with https://nodejs.org/api/dgram.html#dgram_socket_send_msg_offset_length_port_address_callback that if it's not provided or falsy, it will be 127.0.0.1 or ::0 as appropriate.  So what if this was set to an empty string by default, does that work correctly for you?

naseemkullah:
Thanks, I will set `DD_AGENT_HOST` to an empty string and let you know how that goes!

naseemkullah:
However https://github.com/brightcove/hot-shots/blob/5ec98a55c05e0c490ccf993a0a768f5364683fc6/lib/statsd.js#L49 does complicate things as setting the env var to falsey will defaut to using `'localhost'.

naseemkullah:
I'd be happy with changing that line to

```js
 this.host = options.host || process.env.DD_AGENT_HOST; 
```

provided it does indeed default to `127.0.0.1` as we've read and it does not break hot-shots anywhere else.

bdeitte:
@naseemkullah Oh shoot that does make it hard to test.  I'm ok with switching to what you suggest, make sure all tests past, and then marking as a major version to be careful and releasing.  It could be such a nice change, if it's going to improve people's kube and other performance, that it makes sense to me to do this.  Thanks for getting through this with me!

naseemkullah:
That's great news @bdeitte, I look forward to the major release! The fast-path dns lookups for addresses (without additional config) will be a great performance boost when running statsd server on localhost!

test/init.js
```diff
@@ -111,7 +111,7 @@ describe('#init', () => {
 
   it('should set default values when not specified', () => {
     statsd = createHotShotsClient({}, clientType);
-    assert.equal(statsd.host, 'localhost');
```
bdeitte, line 114: Tests are failing now- looks like this needs to now change to undefined

README.md
```diff
@@ -31,7 +31,7 @@ You can check the detailed [change log](https://github.com/brightcove/hot-shots/
 All initialization parameters are optional.
 
 Parameters (specified as one object passed into hot-shots):
-* `host`:        The host to send stats to, if not set, the constructor tries to retrieve it from the `DD_AGENT_HOST` environment variable, `default: localhost`
+* `host`:        The host to send stats to, if not set, the constructor tries to retrieve it from the `DD_AGENT_HOST` environment variable, `default: 127.0.0.1`
```
bdeitte, line 34: Can you change this to indicate that now default: '127.0.0.1' or '::1'

naseemkullah: Sure, applied the change. Should we not set it to `undefined` with a note on dram socket's behaviour?

test/init.js
naseemkullah: Thanks!

naseemkullah:
I've added an out of scope commit which I can revert if you'd like: replacement of deprecated `equal` and `deepEqual` methods.

README.md
bdeitte: That's fine as well, to note it that way certainly.

bdeitte:
I'm perfectly fine with the equal to deepEqual, thanks for the update.  Looks like a few tests need to be updated for this though: https://travis-ci.org/github/brightcove/hot-shots/jobs/728213515#L982

naseemkullah:
> I'm perfectly fine with the equal to deepEqual, thanks for the update. Looks like a few tests need to be updated for this though: https://travis-ci.org/github/brightcove/hot-shots/jobs/728213515#L982

ah yes, thanks!

naseemkullah:
> I'm perfectly fine with the equal to deepEqual, thanks for the update. Looks like a few tests need to be updated for this though: https://travis-ci.org/github/brightcove/hot-shots/jobs/728213515#L982

I've changed a few `undefined`s to `null`s and invoked a `toString()` to a buffer to have the tests pass with `strictEquals` PTAL

bdeitte:
Looks good, thanks for working through this.  I am going to see if one other PR is going to get merged in as well and then will release a major version update with this in it.

naseemkullah:
My pleasure and thanks @bdeitte

PR #187, opened 2020-09-23
Support multiple values for the same tag key

maleblond:
When doing `statsd.increment('someMetricName', 1, undefined, [key:value1, key:value2])`, it currently only send the `key:value2` tag to datadog, ignoring `key:value1`. I believe this is wrong and likely not intended, as DogStatsD supports it (I don't know about Etsy / Telegraf). This PR fixes that so that in the example above, both `key:value1` & `key:value2` will be sent to the StatsD server.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -35,24 +35,27 @@ function overrideTags (parent, child, telegraf) {
     if (idx < 1) { // Not found or first character
       toAppend.push(tag);
     } else {
-      childCopy[tag.substring(0, idx)] = tag.substring(idx + 1);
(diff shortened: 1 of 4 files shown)


bdeitte:
@maleblond Interesting, didn't even know that was a feature.  Looks like a good change to put in, thanks.  I just merged in a PR however which had some test updates in it, which is causing this to need some updates- can you make those?  Thanks again.

maleblond:
:+1: @bdeitte I've just resolved the merge conflicts

bdeitte:
Looks good, will merge in and release a minor version for it

Issue #128, opened 2019-10-07
Handle issues with udsGracefulErrorHandling 


bdeitte:
See https://github.com/brightcove/hot-shots/pull/123#issuecomment-537674855.  It appears we need to handle more error cases for retries.

msiebuhr:
Shouldn't we just try re-connecting on all of them.

And a important detail I noticed: Error-codes are *negative*. My code sees -102, -107 and -111.

bdeitte:
@msiebuhr I see your comments in https://github.com/brightcove/hot-shots/issues/114#issuecomment-540548105 as well so making this a more generic ticket on potential issues with this feature.  Also interesting on what you see for codes, as it's different than what @alessio-modus mentions in https://github.com/brightcove/hot-shots/pull/123

pcothenet:
We're experiencing something like #114. When the agent restart, hot-shots never seem to reconnect properly (error -107)

dhermes:
I'm currently looking into this as well. (I'll probably update this post as I go.) I've got a `local-dd-agent` Docker container / host UDS I've been using to reproduce this.

Error codes:

- 102 / [0x66][3] (Linux): `ENETRESET`
- 107 / [0x6b][4] (Linux): `ENOTCONN`
- 111 / [0x6f][5] (Linux): `ECONNREFUSED`
- 54 / [0x36][2] (macOS): `ECONNRESET`, happens only on the **first** failure when UDS is deleted
- 39 / [0x27][1] (macOS): `EDESTADDRREQ`, happens for **every** attempt to send after deleting the initial UDS (even if a new file replaces it)

[1]: https://github.com/golang/go/blob/go1.15.2/src/syscall/zerrors_darwin_amd64.go#L1157
[2]: https://github.com/golang/go/blob/go1.15.2/src/syscall/zerrors_darwin_amd64.go#L1155
[3]: https://github.com/golang/go/blob/go1.15.2/src/syscall/zerrors_linux_amd64.go#L1272
[4]: https://github.com/golang/go/blob/go1.15.2/src/syscall/zerrors_linux_amd64.go#L1296
[5]: https://github.com/golang/go/blob/go1.15.2/src/syscall/zerrors_linux_amd64.go#L1228

dhermes:
So here is a summary of the problems with the current implementation:

[...]

PR #188, opened 2020-09-25
Fix uds reconnection logic

prognant:
Statsd UDS reconnection logic is not working because error code are checked against positive value : current codebase use the following  positive `errno` : 111 `ECONNREFUSED` & 107 `ENOTCONN`. 

However node.js relies on libuv that uses negated errno :
 * [description in libuv doc](https://docs.libuv.org/en/v1.x/errors.html)
 * [negated errno implementation logic](https://github.com/libuv/libuv/blob/v1.x/include/uv/errno.h#L25-L30)
 * [node.js doc referring to libuv](https://nodejs.org/api/errors.html#errors_error_errno)

This PR uses `util.getSystemErrorName` that was made available in [node.js v9.7.0](https://nodejs.org/api/util.html#util_util_getsystemerrorname_err) and subsequently [backported to v8.12.0](https://nodejs.org/fr/blog/release/v8.12.0/)

Either it's acceptable to bump the node.js version requirements from "Node 8.x and higher" to "Node 8.12 and higher" or we can go back to the hardcoded number and just negate those (side notes: those codes does not exit on mac and are different on windows, -4053 and -4078 thus resolving to the text representation will be more platform agnostic) just let me know !

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -123,33 +123,26 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
     const socketCreateLimit = options.udsGracefulRestartRateLimit || UDS_DEFAULT_GRACEFUL_RESTART_LIMIT; // only recreate once per second
     const lastSocketCreateTime = Date.now();
     this.socket.on('error', (err) => {
-      const code = err.code;
-      switch (code) {
-        case 107:
-        case 111: {
-          if (Date.now() - lastSocketCreateTime >= socketCreateLimit) {
-            // recreate the socket, but only once per 30 seconds
-            if (this.errorHandler) {
-              this.socket.removeListener('error', this.errorHandler);
-            }
-            this.socket.close();
-            this.socket = createTransport(this, {
-              host: this.host,
-              path: options.path,
-              port: this.port,
-              protocol: this.protocol
-            });
-
-            if (this.errorHandler) {
-              this.socket.on('error', this.errorHandler);
-            } else {
-              this.socket.on('error', error => console.error(`hot-shots UDS error: ${error}`));
-            }
+      const code = util.getSystemErrorName(err.code);
+      if (code === 'ENOTCONN' || code === 'ECONNREFUSED') {
+        if (Date.now() - lastSocketCreateTime >= socketCreateLimit) {
+          // recreate the socket, but only once per 30 seconds
+          if (this.errorHandler) {
+            this.socket.removeListener('error', this.errorHandler);
           }
-          break;
-        }
-        default: {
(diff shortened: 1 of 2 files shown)


bdeitte:
@prognant Thank you, well that explains why there were issues here! Two things to note on this:
1. I would really like to still have Node 8 and above, just to not surprise anybody with a lot of usage of this library.  And can note in the code to update later
2. There's another similar PR, funny timing on having these both: https://github.com/brightcove/hot-shots/pull/189/files.  I haven't had a chance yet to figure out yet what makes sense to merge, if they both have different functionality, etc.  So interesting in both your and @dhermes thoughts on this

dhermes:
@bdeitte This PR fixes 1 of the 4 issues I mentioned in https://github.com/brightcove/hot-shots/issues/128#issuecomment-699046970 and is a good start. It could also be done (as a quick hack) by just swapping `107` for `-107` and `111` for `-111`.

Unfortunately a UDS failure with `ENOTCONN` / `ECONNREFUSED` is Linux specific and on macOS those errors are `EDESTADDRREQ` / `ECONNRESET`.

I actually initially had used `util.getSystemErrorName()` but later decided against it because it can throw an error on bad inputs and the error codes are nicely available any how via `os.constants.errno` (I haven't checked how far back they are available, I only checked Node 12 on macOS, Debian and Alpine.)

prognant:
Hi @dhermes & @bdeitte, thanks for your prompt replies.

The PR #189 seems to be a better fit overall. BTW the `os.constants` thing is available at least since node.js 6 as far as I could test, so that's probably the best way not to hardcode errno value.

So I'll close this one and add my thoughts, if any, on the other PR !

Thanks a lot !

dhermes:
@prognant You made some remarks about Windows. Have you done any testing / confirmed how UDS failures look on Windows?

bdeitte:
Thanks @prognant for taking a look here and your thoughts!

prognant:
> @prognant You made some remarks about Windows. Have you done any testing / confirmed how UDS failures look on Windows?

UDS support is very recent on Windows (https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/) and AFAIK datagram are not yet supported. The one thing that comes to my mind regarding Windows is https://github.com/DataDog/datadog-agent/pull/5809 that leverages a Windows specific thing that behaves similarly to UDS. Node.js seems to natively support windows named pipe, but that would probably be a new windows-only feature and the whole UDS thing is likely to remain a linux/bsd/mac thing.

dhermes:
@prognant Awesome info, thanks for the explanation!

PR #189, opened 2020-09-25
Making UDS error handling and recovery more robust.

### LINKED ISSUE #128: Handle issues with udsGracefulErrorHandling

dhermes:
Fixes #128.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1,3 +1,6 @@
+const os = require('os'),
+  process = require('process');
+
 exports.PROTOCOL = {
   TCP: 'tcp',
   UDS: 'uds',
@@ -6,7 +9,22 @@ exports.PROTOCOL = {
 };
 
 function udsErrors() {
-  return [107, 111];
+  if (process.platform == 'linux') {
+    return [
+      os.constants.errno.ENOTCONN,
+      os.constants.errno.ECONNREFUSED,
+    ]
+  }
+
+  if (process.platform == 'darwin') {
+    return [
+      os.constants.errno.EDESTADDRREQ,
+      os.constants.errno.ECONNRESET,
+    ]
+  }
+
+  // Unknown / not yet implemented
+  return []
 };
 
 exports.udsErrors = udsErrors;
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -8,23 +8,28 @@ exports.PROTOCOL = {
   STREAM: 'stream'
 };
 
+/**
+ * Determines error codes that signify a connection to a Unix Domain Socket (UDS)
+ * has failed in a way that can be retried. This codes are OS-specific.
+ * @returns {number[]} An array of the error codes.
+ */
 function udsErrors() {
-  if (process.platform == 'linux') {
+  if (process.platform === 'linux') {
     return [
       os.constants.errno.ENOTCONN,
       os.constants.errno.ECONNREFUSED,
-    ]
+    ];
   }
 
-  if (process.platform == 'darwin') {
+  if (process.platform === 'darwin') {
     return [
       os.constants.errno.EDESTADDRREQ,
       os.constants.errno.ECONNRESET,
-    ]
+    ];
   }
 
   // Unknown / not yet implemented
-  return []
-};
+  return [];
+}
 
 exports.udsErrors = udsErrors;
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -4,3 +4,9 @@ exports.PROTOCOL = {
   UDP: 'udp',
   STREAM: 'stream'
 };
+
+function udsErrors() {
+  return [107, 111];
+};
+
+exports.udsErrors = udsErrors;
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -48,6 +48,8 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   this.cacheDnsTtl = options.cacheDnsTtl || CACHE_DNS_TTL_DEFAULT;
   this.host = options.host || process.env.DD_AGENT_HOST;
   this.port = options.port || parseInt(process.env.DD_DOGSTATSD_PORT, 10) || 8125;
+  this.path = options.path;
+  this.stream = options.stream;
   this.prefix = options.prefix || '';
   this.suffix = options.suffix || '';
   this.tagPrefix = options.tagPrefix || '#';
@@ -68,6 +70,7 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   this.bufferHolder = options.isChild ? options.bufferHolder : { buffer: '' };
   this.errorHandler = options.errorHandler;
   this.udsGracefulErrorHandling = 'udsGracefulErrorHandling' in options ? options.udsGracefulErrorHandling : true;
+  this.udsGracefulRestartRateLimit = options.udsGracefulRestartRateLimit || UDS_DEFAULT_GRACEFUL_RESTART_LIMIT; // only recreate once per second
 
   // If we're mocking the client, create a buffer to record the outgoing calls.
   if (this.mock) {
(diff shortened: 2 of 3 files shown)


dhermes:
Some notes:

- I've been using https://github.com/dhermes/local-dd-agent to start and stop a local `dd-agent`
- I fully expect the maintainers to want more quality control on the (fairly large) change set I dropped from thin air. Please feel free to pick it apart.
- I hope the commits have broken it up in a way that's easy to digest.
- I've added **four** public fields to `Client` (`path`, `stream`, `udsGracefulRestartRateLimit` and `isChild`), this may not be a desirable addition to the "exported" interface
- I haven't updated the TypeScript definition(s) to reflect new exports / fields
- It **may** be considered a breaking change that I've added an attempt to recover in `Client.prototype.sendMessage`

lib/statsd.js
```diff
@@ -526,3 +502,59 @@ Client.prototype.childClient = function (options) {
...
+  }
+
+  client.socket.on('error', (err) => {
+    if (UDS_ERROR_CODES.includes(-err.code)) {
+      udsErrorHandler(client, lastSocketCreateTime);
+    }
+  });
+}
+
+function trySetNewSocket(client) {
```
dhermes, line 576: I'm very open to a naming convention other `try` in `trySetNewSocket()`

lib/statsd.js
```diff
@@ -526,3 +502,59 @@ Client.prototype.childClient = function (options) {
...
+  }
+
+  if (client.errorHandler) {
+    client.socket.on("error", client.errorHandler);
+  } else {
+    client.socket.on("error", (error) => console.error(`hot-shots UDS error: ${error}`));
+  }
+}
+
+function maybeAddUDSErrorHandler(client, lastSocketCreateTime) {
```
dhermes, line 559: I'm very open to a naming convention other than `maybe` in `maybeAddUDSErrorHandler()`

lib/statsd.js
```diff
@@ -526,3 +502,59 @@ Client.prototype.childClient = function (options) {
 
 exports = module.exports = Client;
 exports.StatsD = Client;
+
+function udsErrorHandler(client, lastSocketCreateTime) {
```
dhermes, line 519: 1. I'm not sure if you want these handlers here at the end
1. I'm not sure if you'd prefer `client` be `this` and we just put these helpers on the `Client` prototype

lib/statsd.js
```diff
@@ -354,7 +321,16 @@ Client.prototype.sendMessage = function (message, callback) {
     return;
   }
 
-  if (!this.socket) {
+  const socketWasMissing = !this.socket;
```
dhermes, line 324: It **may** be considered a breaking change that I've added an attempt to recover here

lib/statsd.js
```diff
@@ -48,6 +51,8 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   this.cacheDnsTtl = options.cacheDnsTtl || CACHE_DNS_TTL_DEFAULT;
   this.host = options.host || process.env.DD_AGENT_HOST;
   this.port = options.port || parseInt(process.env.DD_DOGSTATSD_PORT, 10) || 8125;
+  this.path = options.path;
```
dhermes, line 54: - I've added **four** public fields to `Client` (`path`, `stream`, `udsGracefulRestartRateLimit` and `isChild`), this may not be a desirable addition to the "exported" interface
- I haven't updated the TypeScript definition(s) to reflect new exports / fields

dhermes:
To verify this change I've been running `sent-to-uds.js`

```js
const process = require("process");

const hotShots = require("./index");

const DATADOG_ADDRESS = process.env.DATADOG_ADDRESS;

function toPath(datadogAddress) {
  if (datadogAddress.startsWith("unix://")) {
    // NOTE: The length of the "unix://" prefix is 7.
    return datadogAddress.slice(7, datadogAddress.length);
  }

  return datadogAddress;
}

function createUDSClient() {
  const opts = {
    mock: false,
    maxBufferSize: 0,
    path: toPath(DATADOG_ADDRESS),
    protocol: "uds",
    udsGracefulErrorHandling: true,
    udsGracefulRestartRateLimit: 1000,
  };
  return new hotShots.StatsD(opts);
}

function makeEventLoop(udsClient) {
  return function compute() {
    console.log(
      JSON.stringify({
        message: "Doing very big computation",
        level: "info",
        _timestamp: new Date(),
      }),
    );
    udsClient.increment("computation.uds", 1, 1, []);
  };
}

async function main() {
  const udsClient = createUDSClient();
  const compute = makeEventLoop(udsClient);
  setInterval(compute, 500);
}

if (require.main === module) {
  main();
}
```

via

```
SOCKET_DIR="${HOME}/var-run-datadog"
DATADOG_ADDRESS="unix://${SOCKET_DIR}/dsd.socket" node ./sent-to-uds.js
```

and then turning on and off the `local-dd-agent` (https://github.com/dhermes/local-dd-agent) via

```
make run DATADOG_ADDRESS="${SOCKET_DIR}/dsd.socket"
```

bdeitte:
@dhermes Thanks for the PR and all of the info!  Neat to see local-dd-agent too.  I see the tests are still failing in here, looks like something isn't going right on the Mac version.  Also as noted in https://github.com/brightcove/hot-shots/pull/188 looks like we have some overlap here, haven't had a chance yet myself to look over everything and understand.

dhermes:
Yup; I haven't had a chance to parse the test failures but will attempt to tackle it next week. I didn't want to devote too much time building a cathedral before getting sign off from maintainers.

Once / if you think this looks promising I can start to tackle the test failures? I'd love an assist if the Travis CI failures look "easy" to tackle for you (but by all means not needed).

Also, since there are 4 distinct issues, I can send 4 different PRs if you'd rather (the commits themselves in this PR can be broken up nicely to fix each of the 4 issues separately).

lib/constants.js
```diff
@@ -1,6 +1,35 @@
...
+/**
+ * Determines error codes that signify a connection to a Unix Domain Socket (UDS)
+ * has failed in a way that can be retried. This codes are OS-specific.
+ * @returns {number[]} An array of the error codes.
+ */
+function udsErrors() {
+  if (process.platform === 'linux') {
+    return [
+      os.constants.errno.ENOTCONN,
+      os.constants.errno.ECONNREFUSED,
```
bdeitte, line 20: Huh, I didn't realize there was a constant here available on Node 8 and above.  Good to know.

lib/statsd.js
bdeitte: That's fine, and you shouldn't add these to TypeScript definitions, others ones aren't mentioned in there like this (I don't think)

bdeitte:
I took a pretty quick pass on this, but great to see all the work done and makes general sense to me.  When the tests are passing (and I have a little more time) I'll take another look at this and see if I have any more feedback.  This UDS handling has been looking ugly for awhile (from what I can gather from the bug reports), glad to see someone take it all on and make it work better.

I may be able to help with the tests here at some point (do run a Mac here) if you need it, just been short on time recently.  It's not clear to me right away what's going wrong in there- oddly enough it's a mix of all sorts of different tests failing in there (https://travis-ci.org/github/brightcove/hot-shots/jobs/730370815#L1518).

dhermes:
@bdeitte Tests are green here so should be good to review.

lib/statsd.js
bdeitte: Good point, thanks.  Not sure yet if this will be a minor or major release, but I think this part by itself would fit in ok with what has been minor releases.

lib/statsd.js
bdeitte: I think it's fine as is.  Great to see the nice docs there.

lib/statsd.js
bdeitte: I don't get too hung up on names, unless I think it doesn't fit the style.  Not too much of a naming style in here so it's fine.  :)  Spent a bunch of time making sure I understood the new paths in here, for this function and others.  I didn't see any issues, looks good to me.

bdeitte:
Awesome.  I spent some time with tonight and don't see any issues.  Thanks again for this!  Going to release in a minor release here in a bit.

PR #191, opened 2020-11-13
Add gaugeDelta function

Dinoshauer:
Hi! We had the need to use a gauge with incremental og decremental deltas the other day, I found out `hot-shots` do not have support for it, but it can be done with the `#send` method, so I figured I would add a pull request for the feature, it's based on how [statsd-client does it](https://github.com/msiebuhr/node-statsd-client/blob/master/lib/statsd-client.js#L63) (well heavily copied :)).

However, I am not sure how to write the tests for this function. I could add another key to the statFunction test array, perhaps like this:

```js
{ name: 'gaugeDelta', unit: 'g', bytes: 12, prefix: '+' },
{ name: 'gaugeDelta', unit: 'g', bytes: 12, prefix: '-' }
```

What do you think? I wanted to get your input before going ahead and changing the tests

### DIFF
--- a/lib/statsFunctions.js
+++ b/lib/statsFunctions.js
@@ -187,7 +187,6 @@ function applyStatsFns (Client) {
     this.sendAll(stat, value, 'd', sampleRate, tags, callback);
   };
 
-
   /**
    * Gauges a stat by a specified amount
    * @param stat {String|Array} The stat(s) to send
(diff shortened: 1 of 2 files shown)


bdeitte:
Oh I didn't realize this was a thing, nice to see it's added.  Yes that addition to the tests sounds good to me.  Please also update the README with an example and if you can, update types.d.ts.  Thanks!

Dinoshauer:
Sure thing, I'll see if I can get around to it this week or the next

bdeitte:
@Dinoshauer Just wanted to check to see if this is something you were still going to do.  Thanks!

Dinoshauer:
Hi @bdeitte - sorry for the radio silence! I don't think I will have time to implement the tests in the near future, so if you or someone else wants to take it over, please do!

bdeitte:
Ok thanks for the heads up.  Closing this PR then and opening an issue on it https://github.com/brightcove/hot-shots/issues/200

PR #193, opened 2020-11-23
always emit error on socket after unsuccessful send

ralphiech:
Emit error after send failure in all cases (callback, custom error handler, default error handler) as opposed to only on default error handler

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -354,9 +354,11 @@ Client.prototype.sendMessage = function (message, callback) {
         this.errorHandler(errFormatted);
       } else {
         console.error(String(errFormatted));
-        // emit error ourselves on the socket for backwards compatibility
-        this.socket.emit('error', errFormatted);
       }
+    }
+    if (errFormatted) {
+      // emit error ourselves on the socket for backwards compatibility
+      this.socket.emit('error', errFormatted);
     }
   };
 


bdeitte:
Hi @ralphiech thanks for the PR.  I believe this will break existing use cases- right now the code is saying "if I don't have any other way to handle it that passed in, emit an error", while you would like it to always emit an error no matter what was passed in.  A test breaks for this case, which indicates to me it will break some people's apps.  Can you describe what you need this for?  What about a option passed in to enable this behavior?

bdeitte:
@ralphiech closing out, but please feel free to reopen if there's a way to do this that and keep compatibility, thanks.

ralphiech:
Hey @bdeitte sorry for late reply. I read the code different. It still only emits an error in case of an error (`if (errFormatted) {` is still there). What is different is that this behaviour now is consistent no matter if an error handler function is provided or not, which in my opinion is correct or do you disagree?

bdeitte:
Hi @ralphiech so the inconsistency you are seeing has always been built into things, which is what I'm worried about breaking given it's been there from the very beginning.  I strive for backwards comaptibilty in almost all cases with this library, and doing this is going to break someone.  It's not exactly this case, but the README speaks to something similar: "If the optional callback is not given, an error is thrown in some cases and a console.log message is used in others. An error will only be explicitly thrown when there is a missing callback or if it is some potential configuration issue to be fixed."

ralphiech:
Hey @bdeitte 

I understand that you do not want to break backwards compatibility, however are people actually actioning on error events, usually its all done via explicitly called error handler right? However even if people are using events, my PR would in fact flag up a real issue, which in the past, in my opinion, was not raised wrongly (in that sense fix a bug).

The side effect of this is that the uds error handler without the PR is only working reliably, if no custom error handler is set. This again seems wrong to me as the lib supports custom error handler as well as `udsGracefulErrorHandling` at the same time.

PR #194, opened 2020-12-01
Make close callback optional in TypeScript definition

stephenmathieson:
This patch makes the callback to `StatsD#close()` optional in the TypeScript definitions.


bdeitte:
Makes sense, thanks.

stephenmathieson:
Thanks for the quick merge!

PR #195, opened 2020-12-04
Handle UDS errors occurring when sending metrics

chotiwat:
The UDS error handling added in #189 only triggers when an `error` event is emitted on the current socket instance. When the client encounters an error from sending a message, it does not send an error event to the underlying socket if the callback or the `errorHandler` option is supplied. I have been able to reproduce the issue with the following steps.

1. Start up a DogStatsD server that listens on a UDS.  (I used https://github.com/dhermes/local-dd-agent.)
2. Point the `hot-shots` client to the socket path with a custom `errorHandler`. Periodically sending a metric with `setInterval`.
3. Stop the server.
4. The client should start triggering the `errorHandler` with a UDS error.
5. Start the server which would recreates the socket file.
6. The client cannot reconnect to the socket path and sending metrics still triggers the `errorHandler`. I expect the client to be able to reconnect after the DogStatsD server is back online.

This PR adds UDS error handling for such case by calling `udsErrorHandler()` directly. The `lastSocketCreateTime` field is added to the client to allow calling the function outside the scope that the socket is created.

I also added two test cases to represent the issue this PR is trying to solve.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -345,6 +345,8 @@ Client.prototype.sendMessage = function (message, callback) {
     const errFormatted = err ? new Error(`Error sending hot-shots message: ${err}`) : null;
     if (errFormatted) {
       errFormatted.code = err.code;
+      // handle UDS error that requires socket replacement when we are not
+      // emitting the `error` event on `this.socket`
       if (this.protocol === PROTOCOL.UDS && (callback || this.errorHandler)) {
         udsErrorHandler(this, err);
       }
@@ -506,7 +508,7 @@ exports = module.exports = Client;
 exports.StatsD = Client;
 
 /**
- * Handle an error connecting to a Unix Domain Socket (UDS). This will
+ * Detect and handle an error connecting to a Unix Domain Socket (UDS). This will
  * attempt to create a new socket and replace and close the client's current
  * socket, registering a **new** `udsErrorHandler()` on the newly created socket.
  * If a new socket can't be created (e.g. if no UDS currently exists at
@@ -571,7 +573,8 @@ function maybeAddUDSErrorHandler(client) {
 
 /**
  * Try to replace a client's socket with a new transport. If `createTransport()`
- * returns `null` this will still set the client's socket to `null`.
+ * returns `null` this will still set the client's socket to `null`. This also
+ * updates the socket creation time for UDS error handling.
  * @param client Client The statsd Client that will be getting a new socket
  */
 function trySetNewSocket(client) {
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -525,8 +525,8 @@ function udsErrorHandler(client, err) {
   }
 
   // recreate the socket, but only once within `udsGracefulRestartRateLimit`.
-  if (!client.lastSocketCreateTime ||
-    Date.now() - client.lastSocketCreateTime < client.udsGracefulRestartRateLimit) {
+  if (!client.socket.createdAt ||
+    Date.now() - client.socket.createdAt < client.udsGracefulRestartRateLimit) {
     return;
   }
 
@@ -543,7 +543,6 @@ function udsErrorHandler(client, err) {
   if (newSocket) {
     client.socket.close();
     client.socket = newSocket;
-    client.lastSocketCreateTime = Date.now();
     maybeAddUDSErrorHandler(client);
   } else {
     console.error('Could not replace UDS connection with new socket');
@@ -589,5 +588,4 @@ function trySetNewSocket(client) {
     protocol: client.protocol,
     stream: client.stream,
   });
-  client.lastSocketCreateTime = Date.now();
 }
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -63,7 +63,7 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   if (process.env.DD_ENTITY_ID) {
     this.globalTags = this.globalTags.filter((item) => {
       return item.indexOf('dd.internal.entity_id:') !== 0;
-   });
+    });
     this.globalTags.push('dd.internal.entity_id:'.concat(helpers.sanitizeTags(process.env.DD_ENTITY_ID)));
   }
   this.telegraf = options.telegraf || false;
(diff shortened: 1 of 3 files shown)


test/errorHandling.js
```diff
@@ -181,140 +181,207 @@ describe('#errorHandling', () => {
...
-            }), 'client');
-            const initialSocket = client.socket;
-            setTimeout(() => {
-              initialSocket.emit('error', { code });
-              assert.ok(Object.is(initialSocket, client.socket));
-              // it should not create the socket anyway if it breaks too quickly
-              // change time and make another error
-              Date.now = () => 4857394578 + 1000; // 1 second later
-              initialSocket.emit('error', { code });
+          it('should re-create the socket on bad descriptor error when sending metric with a callback', (done) => {
```
chotiwat, line 320: These are the two new test cases.

test/errorHandling.js
```diff
@@ -181,140 +181,207 @@ describe('#errorHandling', () => {
...
-            setTimeout(() => {
-              initialSocket.emit('error', { code });
-              assert.ok(Object.is(initialSocket, client.socket));
-              // it should not create the socket if it breaks too quickly
-              // change time and make another error
-              Date.now = () => 4857394578 + 1000; // 1 second later
-              initialSocket.emit('error', { code });
+          afterEach(() => {
+            Date.now = realDateNow;
+          })
```
chotiwat, line 190: When trying to confirm that my new test cases break without this PR's change, I found that other timing related tests broke, most likely because `Date.now` wasn't properly restored after an assertion error. Using `afterEach` within a sub-`describe` block ensures that we always restore `Date.now` to the original.

lib/statsd.js
```diff
@@ -583,4 +587,5 @@ function trySetNewSocket(client) {
     protocol: client.protocol,
     stream: client.stream,
   });
+  client.lastSocketCreateTime = Date.now();
```
chotiwat, line 590: Should I move this into `createTransport`?

lib/statsd.js
```diff
@@ -346,6 +345,11 @@ Client.prototype.sendMessage = function (message, callback) {
     const errFormatted = err ? new Error(`Error sending hot-shots message: ${err}`) : null;
     if (errFormatted) {
       errFormatted.code = err.code;
+      // handle UDS error that requires socket replacement when we are not
+      // emitting the `error` event on `this.socket`
+      if (this.protocol === PROTOCOL.UDS && (callback || this.errorHandler)) {
+        udsErrorHandler(this, err);
+      }
```
dhermes, line 352: If I understand correctly this is because the code below does the following:

```js
if (callback) {
  callback(errFormatted, bytes);
} else if (err && this.errorHandler) {
  this.errorHandler(errFormatted);
} else if (err) {
  console.error(String(errFormatted));
  this.socket.emit('error', errFormatted);
}
```

i.e. the only case that the `error` event is emitted (when `err` is truth-y) is when `callback` and `errorHandler` are false-y?

Does it not make sense (it's a little repetitive) to `this.socket.emit()` in all cases?

lib/statsd.js
```diff
@@ -513,12 +517,14 @@ exports.StatsD = Client;
  * Note that this will no-op with an early exit if the last socket create time
  * was too recent (within the UDS graceful restart rate limit).
  * @param client Client The statsd Client that may be getting a UDS error handler.
```
dhermes, line 519: Make sure to update the JSDoc to refer to `err`

lib/statsd.js
```diff
@@ -535,7 +541,8 @@ function udsErrorHandler(client, lastSocketCreateTime) {
   if (newSocket) {
     client.socket.close();
     client.socket = newSocket;
-    maybeAddUDSErrorHandler(client, Date.now());
+    client.lastSocketCreateTime = Date.now();
```
dhermes, line 544: IMO it's strange to bolt random things onto this type. But maybe this isn't so random.

lib/statsd.js
```diff
@@ -513,12 +517,14 @@ exports.StatsD = Client;
...
  */
-function udsErrorHandler(client, lastSocketCreateTime) {
+function udsErrorHandler(client, err) {
+  if (!err || !UDS_ERROR_CODES.includes(-err.code)) {
+    return;
+  }
+
   // recreate the socket, but only once within `udsGracefulRestartRateLimit`.
-  if (Date.now() - lastSocketCreateTime < client.udsGracefulRestartRateLimit) {
+  if (Date.now() - client.lastSocketCreateTime < client.udsGracefulRestartRateLimit) {
```
dhermes, line 527: Don't you need to account for code paths where `client.lastSocketCreateTime` is not set?

lib/statsd.js
dhermes: IMO it's critical (pending discussion of whether it makes sense to bloat the interface of `client`)

lib/statsd.js
chotiwat: Exactly. I tried emitting the error on the socket in all cases at first but decided against it because of this comment:
https://github.com/chotiwat/hot-shots/blob/450715ecc308addd74640b514e2a78330a194a48/lib/statsd.js#L361-L362

lib/statsd.js
chotiwat: I actually meant moving the assignment of `Date.now()`, but maybe we can move the field to `client.socket` and set it in `createTransport` like we do the `type` field.
https://github.com/brightcove/hot-shots/blob/a64057a6c4f02b5ed38ddd4249854571f0412864/lib/transport.js#L160

lib/statsd.js
chotiwat: I needed somewhere to cache the socket creation time because we didn't have access to it in `Client.prototype.sendMessage`. I'm open to move it elsewhere (currently leaning towards `client.socket` https://github.com/brightcove/hot-shots/pull/195/files#r537880161).

bdeitte:
Thanks @chotiwat, and thanks @dhermes for the great review.  It took me a bit longer than usual to get to this (and then took me awhile to make sure I understood it correctly), but this all now looks good to me.  Going to merge in and make a minor release for it.

PR #201, opened 2021-04-01
Fix formatDate ignoring number primitives

dvd-z:
`formatDate` currently returns `undefined` when you pass it a primitive number since `1 instanceof Number === false`. This will fix it so it detects both Number objects and primitive numbers.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -68,7 +68,7 @@ function formatDate(date) {
   if (date instanceof Date) {
     // Datadog expects seconds.
     timestamp = Math.round(date.getTime() / 1000);
-  } else if (date instanceof Number) {
+  } else if (date instanceof Number || typeof date === 'number') {
     // Make sure it is an integer, not a float.
     timestamp = Math.round(date);
   }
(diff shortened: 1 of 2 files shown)


bdeitte:
@dvd-z Thanks!  Could you add a test for this case?  Something like this in test/event.js should work:

```
      it('should send proper event format for title, text, and options', done => {
        const date = 1;
        const dateAsDate = new Date(date);
        server = createServer(serverType, opts => {
          statsd = createHotShotsClient(opts, clientType);
          const options = {
            date_happened: date,
            hostname: 'host',
            aggregation_key: 'ag_key',
            priority: 'low',
            source_type_name: 'source_type',
            alert_type: 'warning'
          };
          statsd.event('test title', 'another\nmultiline\ndescription', options);
        });
        server.on('metrics', event => {
          assert.strictEqual(event, `_e{10,31}:test title|another\\nmultiline\\ndescription|d:${Math.round(dateAsDate.getTime() / 1000)}|h:host|k:ag_key|p:low|s:source_type|t:warning${metricEnd}`
          );
          done();
        });
      });

```

dvd-z:
Pushed! I did change it a little since I believe `date_happened` is seconds when passed in as a number.

PR #206, opened 2021-05-25
Fixes #205 catch error when trying to connect with UDS and close socket

### LINKED ISSUE #205: Unix domain socket is not closed after unsuccessfully attempting to connect
[The call to connect on a UDS socket](https://github.com/brightcove/hot-shots/blob/2fd413630eab649274f301ab13f9ba5c4cd64abb/lib/transport.js#L102) can raise an error if the socket is not available.  This results in a leak and these sockets are never cleaned up. A simple fix would be wrap the call to `connect` in a try/catch and close the socket upon catching an error. To keep the behaviour consistent with current functionality we can rethrow the error after closing the socket.

If there are no objections to this approach I'm happy to submit a PR with the changes.

cmaddalozzo:


bdeitte:
Thanks for the fix and explanation in the ticket- makes sense to me.

PR #207, opened 2021-05-31
Fix 170

### LINKED ISSUE #170: TCP protocol does not try to reconnect
I suppose this isn't intended given that uds seems to have such feature? When I restart telegraf instance, hot-shots never tries to make a new connection.  It just keeps getting the same error over and over again

```
Error: This socket has been ended by the other party
    at Socket.writeAfterFIN [as write] (net.js:451:14)
    at Object.send (C:\Users\User\monitor_test\node_modules\hot-shots\lib\transport.js:36:14)
    at Client.sendMessage (C:\Users\User\monitor_test\node_modules\hot-shots\lib\statsd.js:385:17)
    at Client._send (C:\Users\User\monitor_test\node_modules\hot-shots\lib\statsd.js:300:12)
    at Client.send (C:\Users\User\monitor_test\node_modules\hot-shots\lib\statsd.js:277:8)
    at Client.sendStat (C:\Users\User\monitor_test\node_modules\hot-shots\lib\statsd.js:254:8)
    at Client.sendAll (C:\Users\User\monitor_test\node_modules\hot-shots\lib\statsd.js:230:10)
    at Client.increment (C:\Users\User\monitor_test\node_modules\hot-shots\lib\statsFunctions.js:114:10)
    at asd (C:\Users\User\monitor_test\src\index.js:17:20) {
  code: 'EPIPE'
}
```

cesarfd:
Fixes #170 

Hi, this is an attempt to make the library use the same logic implemented for UDS to re-connect a socket but also for TCP. It seems to work fine but I'm not sure I've followed the right approach here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ Parameters (specified as one object passed into hot-shots):
 * `protocol`: Use `tcp` option for TCP protocol, or `uds` for the Unix Domain Socket protocol or `stream` for the raw stream. Defaults to `udp` otherwise.
 * `path`: Used only when the protocol is `uds`. Defaults to `/var/run/datadog/dsd.socket`.
 * `stream`: Reference to a stream instance. Used only when the protocol is `stream`.
+* `tcpGracefulErrorHandling`: Used only when the protocol is `tcp`. Boolean indicating whether to handle socket errors gracefully. Defaults to true.
+* `tcpGracefulRestartRateLimit`: Used only when the protocol is `tcp`. Time (ms) between re-creating the socket. Defaults to `1000`.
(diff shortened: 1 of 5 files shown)


cesarfd:
No idea how to make it work in windows platforms. I've added equivalent? `os.errno` codes but still fails.

bdeitte:
Hi @cesarfd sorry for the delay in a review.  This approach looks fine to me, but I did not review in depth.  I am not sure what is wrong in the tests offhand.

bdeitte:
This looks good- thanks for all the work here!

Given this changes a default behavior, and given how much trouble the general areas has given in the past, bumping the major version for this change.

PR #209, opened 2021-07-15
Add a closingFlushInterval option which allows stopping StatsD quicker

maxday:
👋  Team !

When `close` is called, we need to wait at least 50ms. This could be quite long in a time-constraint environment such as an AWS Lambda function.

This PR adds a simple option to customize this parameter, defaulting to the current value : 50ms which is fully backward-compatible.

Let me know what do you think! 🙏


bdeitte:
@maxday This looks like a great addition and makes sense to me.  Merging in and will make a minor release for it.  Thanks!

PR #212, opened 2021-08-30
fix(types): add stream to protocol string union type

### LINKED ISSUE #211: Update type definitions to encapsulate stream protocol
String union type definition for [protocol option](https://github.com/brightcove/hot-shots/blob/master/types.d.ts#L20) does not include `stream`. Will raise a PR to fix this shortly.

tim-crisp:
Fixes #211.

Adds `stream` to protocol string union type.


bdeitte:
Looks good- thanks for the fix.

PR #213, opened 2021-09-15
Add missing decrement overload type

amc6:
It appears that the code already supports this overload, so I assume that it was just accidentally left out of the type definition. https://github.com/amc6/hot-shots/blob/master/lib/statsFunctions.js#L149


bdeitte:
Yes, looks like it, thanks for the PR

Issue #215, opened 2021-10-20
Prefix (and suffix?) should probably enforce separator presence


sambostock:
## TL;DR

We should enforce that metric prefixes end in a period, or insert it automatically. 

## Reasoning

Consider the following metric:

```
request.time
```

It is common to want to add a prefix to the metric to identify the application emitting it.

In many other StatsD libraries, specifying a non-empty prefix (e.g. `my_app`) results in the following:

```
my_app.request.time
```

A period (`.`) is automatically added to delimit the metric name from the prefix, unless it is empty.

However, this library produces the following:

```diff
-my_app.request.time
+my_apprequest.time
```

To get this behaviour with `hot-shots`, one must specify the separator as part of the prefix: `my_app.`. This is error prone, especially as it does not appear to be documented.

I am not aware of scenarios in which one would want a prefix to be joined to the metric name without a separator, though perhaps they exist.

**If there are no such use cases, I think it would be sensible to either**
- **require the last character of a non-empty prefix to be a period, or**
- **automatically add a period in between the prefix and metric name if one is not already present.**

This may also apply to metric suffixes, although of course the first character would be considered instead of the last.

tlhunter:
`hot-shots` should absolutely insert the period when missing. I get bit by this with every new project ;)

bdeitte:
I'm totally good with a PR for this if anybody wants to create one.  This will require a major version update, just in case it would break someone, but that does seem like something that just trips everyone up and is never needed.

bdeitte:
And apologies (as I keep saying as I get through old issues here) on the delay on this.

PR #216, opened 2021-10-21
Document explicit prefix/suffix separators

sambostock:
This adds documentation for the `prefix` and `suffix` configs.

No separator (e.g. `.`) is added, so it should be included in the given string to separate the prefix/suffix from the rest of the metric name. See #215 for more context.

I tried implementing auto-separating prefixes and suffixes, but it turns out to be a little complex to keep the option for the old behaviour. This at least clarifies the existing behaviour, until such as time as it is decided to change it.


bdeitte:
Sorry for the delay in review, and thanks for the info on this!

PR #220, opened 2022-02-21
Fix tests

dipunm:
Ran tests on a Mac M1 using node 14.

Got error: 
```
 Uncaught AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 'a:1|c\nb:2|c\n'
- 'a:1|c\n'
          ^
      + expected - actual

       a:1|c
      -b:2|c
```

Looks like the number of characters for the first message is only 5 characters long so I suspect this is a broken test, however without more context this could actually be exposing a broken feature in which case this PR can be closed.

### DIFF
--- a/test/buffer.js
+++ b/test/buffer.js
@@ -65,7 +65,7 @@ describe('#buffer', () => {
       it('should not send batches larger then maxBufferSize', done => {
         server = createServer(serverType, opts => {
           statsd = createHotShotsClient(Object.assign(opts, {
-            maxBufferSize: 8,
+            maxBufferSize: 5,
           }), clientType);
           statsd.increment('a', 1);
           statsd.increment('b', 2);


dipunm:
5 is the exact number of characters, and it looks like it flushes the buffer correctly meaning it proves that the buffer flushes once its limit is reached, however if we don't want to test edge cases, we could use the value 4 which makes the test become: the buffer flushes as soon as its limit is exceeded.

dipunm:
In hindsight, after reading the test again, it looks like maybe this is a broken feature. The buffer should not have flushed the messages together.

But this contradicts the documentation:
```
* `maxBufferSize`: If larger than 0,  metrics will be buffered and only sent when the string length is greater than the size. `default: 0`
```

I will leave the PR as is so this can be seen and then the solution can be decided.

dipunm:
After switching to node 16 and re-installing dependencies, the tests started passing.
I then switched to node 14 and re-installed dependencies and the tests are still passing.

This is now confusing me but I cannot reproduce the failing tests anymore.

Leaving this open just to keep visibility and to allow author to see the contradiction between the test and the readme.

dipunm:
Immediately after writing that comment, it failed again using node 14. I think the issue is intermittent. This might have been better as an issue 😬 

I have been running `npm it` repeatedly and get a pass more than a fail but still see fails. I have seen it pass and fail.
I also managed to get it failing with node 16. (not the difference between node 14 and node 16 is that node 14 runs in Rosetta emulation mode) and node 16 is running using native arm binaries.

bdeitte:
Hi, thanks for the update here, and sorry for the wait on this. I've been very responsive in the course of this project, until the last six months, when I have not been able to give it any attention. I do have some time today and will be looking to merge this.

bdeitte:
Ok so now reading through this, I'm also confused!  :)  Not sure what is going on here either.  Please feel free to create an issue and let me know if you figure out anything more here.

Issue #218, opened 2022-02-08
Client closing function failing when errorHandler is defined, but socket is null.


dlancer:
Client closing function failing when errorHandler is defined, but socket is null,  e.g. uds transport failed to create for some reason.

```
TypeError: Cannot read property 'removeListener' of null
^
this.socket.removeListener('error', this.errorHandler);
```
https://github.com/brightcove/hot-shots/blob/master/lib/statsd.js#L452

lizzie:
got same errors  @bdeitte 

```
04/25/2022, 11:11:52  [Unknown] [ERROR] uncaughtException: Cannot read properties of null (reading 'removeListener')
TypeError: Cannot read properties of null (reading 'removeListener')
    at ChildClient.Client._close (/Users/yan.sheng/workspace/modustech/notarize/node_modules/hot-shots/lib/statsd.js:453:17)
    at Timeout.<anonymous> (/Users/yan.sheng/workspace/modustech/notarize/node_modules/hot-shots/lib/statsd.js:419:14)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7)
```

batbrain9392:
Hi! Was wondering if this has been fixed (https://github.com/brightcove/hot-shots/pull/227), maybe it can be closed?

batbrain9392:
Unfortunately, we're still facing it in v10.

```bash
TypeError Cannot read properties of null (reading 'removeListener')
  at Client._close(/app/node_modules/hot-shots/lib/statsd.js:453:16)
    this.socket.removeListener('error', this.errorHandler);
  at Timeout.<anonymous>(/app/node_modules/hot-shots/lib/statsd.js:419:13)
    this._close(callback);
  at listOnTimeout(node:internal/timers:569:16)
  at process.processTimers(node:internal/timers:512:6)
```

bdeitte:
CHANGELOG is available and updates continue to be made. Let me know if you continue to have issues here.

PR #227, opened 2022-06-20
Check if client is undefined to fix errors

bdeitte:
Fix for #218

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 ## 9.1.0 (2022-6-20)
 * @bdeitte Update testing dependencies
+* @bdeitte Check if client is undefined before closing to fix error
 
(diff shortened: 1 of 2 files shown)


batbrain9392:
Unfortunately, we're still facing it in v10.

```bash
TypeError Cannot read properties of null (reading 'removeListener')
  at Client._close(/app/node_modules/hot-shots/lib/statsd.js:453:16)
    this.socket.removeListener('error', this.errorHandler);
  at Timeout.<anonymous>(/app/node_modules/hot-shots/lib/statsd.js:419:13)
    this._close(callback);
  at listOnTimeout(node:internal/timers:569:16)
  at process.processTimers(node:internal/timers:512:6)
```

batbrain9392:
https://github.com/brightcove/hot-shots/issues/218#issuecomment-1655807870

PR #231, opened 2022-07-21
Add udpSocketOptions as a StatsD option

### LINKED ISSUE #223: Cache DNS implementation is susceptible to a stampede
First, thank you for making this client. It has been very useful.

At $JOB, we have a load bearing service sending stats via this client. One challenge is that we are on k8s, which has some real problems. See [Racy conntrack and DNS lookup timeouts](https://www.weave.works/blog/racy-conntrack-and-dns-lookup-timeouts). Ideally, we should use [NodeLocal DNSCache](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/) but for $REASONS we have not made that change.

We have seen [dns.lookup](https://nodejs.org/api/dns.html#dnslookuphostname-options-callback) take multiple  seconds due to this. Our collector rarely changes its IP address, so caching the results is a natural solution.

Recently, I was noticing hundreds of calls to `dns.lookup` in our traces and tracked it down to:

https://github.com/brightcove/hot-shots/blob/02f26a6d5da6fd2351404afc39c45836a1f017f5/lib/transport.js#L56-L71

Let us assume:

- the service is sending one stat every 1ms
- `dns.lookup` takes 100ms to respond

When we cache miss, we will make roughly 100 calls to `dns.lookup` before one call succeeds and we updated `dnsResolutionData` with the results. When combined with the above k8s DNS issues, each subsequent call to `dns.lookup` exacerbates the issue and we balloon to multi-second DNS lookups.

The solution I have implemented internally uses the stale IP address until the next DNS lookup succeeds. Would you be open to such a patch? My use case is a bit unique, but it does seem like folks may want to opt-in to that behavior as the `dns.lookup` call is not really asynchronous.

> DNS requests in node appear asynchronous, but they're actually internally implemented as synchronous calls within node's internal libuv threadpool (which by default has only 4 threads). That means if you do >4 DNS lookups in parallel then you're going to block the libuv threadpool

Source: https://httptoolkit.tech/blog/configuring-nodejs-dns/

hjr3:
This options allows the library user to control how the UDP socket is
created. If not specified, we maintain existing compatibility by
defaulting to a udp4 socket family with no other options.

Fixes #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ Parameters (specified as one object passed into hot-shots):
 * `udsGracefulErrorHandling`: Used only when the protocol is `uds`. Boolean indicating whether to handle socket errors gracefully. Defaults to true.
 * `udsGracefulRestartRateLimit`: Used only when the protocol is `uds`. Time (ms) between re-creating the socket. Defaults to `1000`.
 * `closingFlushInterval`: Before closing, StatsD will check for inflight messages. Time (ms) between each check. Defaults to `50`.
+* `udpSocketOptions`: Specify these options to control how StatsD will create a UDP (dgram) socket.
 
 ### StatsD methods
 All StatsD methods other than `event`, `close`, and `check` have the same API:
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -87,6 +87,7 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   this.udsGracefulRestartRateLimit = options.udsGracefulRestartRateLimit || UDS_DEFAULT_GRACEFUL_RESTART_LIMIT; // only recreate once per second
   this.isChild = options.isChild;
   this.closingFlushInterval = options.closingFlushInterval || 50;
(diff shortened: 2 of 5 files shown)


eladchen (approved):

README.md
```diff
@@ -65,6 +65,7 @@ Parameters (specified as one object passed into hot-shots):
 * `udsGracefulErrorHandling`: Used only when the protocol is `uds`. Boolean indicating whether to handle socket errors gracefully. Defaults to true.
 * `udsGracefulRestartRateLimit`: Used only when the protocol is `uds`. Time (ms) between re-creating the socket. Defaults to `1000`.
 * `closingFlushInterval`: Before closing, StatsD will check for inflight messages. Time (ms) between each check. Defaults to `50`.
+* `udpSocketOptions`: Specify these options to control how StatsD will create a UDP (dgram) socket.
```
bdeitte, line 68: I'm going to tweak the wording of this a little bit, to make clear for those who don't know about it what "these" refers to.  But I can do that myself after merging, so I can just get this all set now.

Issue #233, opened 2022-08-24
StatsD ClientOptions interface is missing `stream` property


albert-mirzoyan:
The runtime supports the `stream` option
https://github.com/brightcove/hot-shots/blob/master/lib/statsd.js#L64
But the types.d.ts is missing the declaration.
https://github.com/brightcove/hot-shots/blob/master/types.d.ts#L5

bdeitte:
Thanks for the heads up.  PRs are always welcome, or I'll try to remember this the next time I'm merging things and making updates.

bdeitte:
I am making a small update myself right now, and I was going to put this in.  But I'm not using TypeScript and wasn't 100% sure the type should  be Stream.  So until next time I make a change, or someone else can submit the PR.

PR #236, opened 2022-10-22
add stream property type to ClientOptions (#233)

albert-mirzoyan:


bdeitte:
Thanks.  I have another PR to make shortly, and then I will merge both in tomorrow and make a release.

PR #245, opened 2023-01-30
Update Node versions to use and some small README improvements

bdeitte:


bdeitte:
Hmmm, so this is failing locally I can now see.  For the TCP test for "should aggregate packets when maxBufferSize is set to non-zero", I get this locally: "Error: Error sending hot-shots message: Error [ERR_STREAM_DESTROYED]: Cannot call write after a stream was destroyed".  And other tests also fail like this, but focusing on this one first.  I think that the tests are now too quick, something that likely doesn't come up in the real world but happen in how these are constructed.  I have a feeling we need to wait for the socket 'connect' event somewhere, but I haven't figured out yet where would be best.

PR #224, opened 2022-05-10
append standard Datadog tags from env vars

zhyu:
`env`, `service` and `version` are standard Datadog tags: https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging/?tab=kubernetes

Since the `dd.internal.entity_id` tag has already been set from `DD_ENTITY_ID` env var, I think extending it to more standard tags makes sense.

Actually, the same thing has been done in the standard [Golang library](https://github.com/DataDog/datadog-go/blob/553de96e699a42be8b401607fbbbce81d4942790/statsd/statsd.go#L89-L98) and [Java library](https://github.com/DataDog/java-dogstatsd-client/pull/107) as well.

PR #250, opened 2023-05-10
Only add DD_ prefixed tags if mapped tag value is not specified

howinator:
#224 is great, but I'm currently blocked on upgrading hot-shots because, for some of our services, we specify a value for `service` which is different than the value in the `DD_SERVICE` env var.

This PR modifies the logic so that user-provided globalTags take precedence over DD_ prefix tags. I believe this to be the least surprising behavior, but let me know what you think.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -69,7 +69,7 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   this.mock        = options.mock;
   this.globalTags  = typeof options.globalTags === 'object' ?
       helpers.formatTags(options.globalTags, options.telegraf) : [];
-  const availableDDEnvs = Object.keys(DD_ENV_GLOBAL_TAGS_MAPPING).filter(key => process.env[key]);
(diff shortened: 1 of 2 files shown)


chris-heap (approved):

greghuels (approved):

bdeitte:
Apologies for a slow response here, and thanks for the PR.  This feels to me as well as a reasonable suggestion and not something at all likely to break backward compatibility (always a big concern for this library).  Merging it in, and I'll make a minor release for it.

bdeitte:
Actually, as I really review, I'm not sure about the backwards compatibility and merging in.  I was assuming there was a bug in what was there before I reviewed the code, that when DD_ tags were there along with globalTags, that there was an issue.  Now I'm understanding that this just reverses the order of what takes priority.  I don't love changing that after it's released, and I would be interested in @zhyu or other's opinions on this.

zhyu:
Well, it's not backward compatible. And if we're going to merge it in, at least we should update the README about this overriding behavior.

bdeitte:
Closing out given the time opened and not wanting to change backwards compatibility by default.

Issue #200, opened 2021-03-30
Add gaugeDelta function


bdeitte:
See https://github.com/brightcove/hot-shots/pull/191 which just needs to be finished up

stepanho:
Hello @bdeitte! Can we escalade resolving that issue?

bdeitte:
Hi @stepanho this project is all based on people contributing at this point.  So if you or someone else would like to finish up that work, I am happy to review and merge in.

PR #259, opened 2023-10-13
Add gaugeDelta function

lachlankidson:
A reraise of https://github.com/brightcove/hot-shots/pull/191 (with a minor correction to prevent double negative symbols) and added tests + types. Addresses https://github.com/brightcove/hot-shots/issues/200


bdeitte:
Thanks for the contribution, and apologies for the long delay here.  I'm the sole person reviewing things, and life got in the way of handling things here for awhile.  This looks good to me, and I'll cut a release with it in it after looking at other outstanding items.

PR #271, opened 2024-09-09
Handle synchronous socket.send error in sendUsingDnsCache

matteosb:
We are using UDP with DNS caching in production and noticed the occasional unhandled exception in our application when the datadog agent is unavailable. We tracked it down to the lookup branch of `sendUsingDnsCache` which calls `socket.send` within a callback. `socket.send` can throw a sync exception and not invoke `callback` as described here: https://nodejs.org/api/dgram.html#socketsendmsg-offset-length-port-address-callback.

### DIFF
--- a/lib/transport.js
+++ b/lib/transport.js
@@ -63,7 +63,11 @@ const createUdpTransport = args => {
         }
         dnsResolutionData.resolvedAddress = address;
         dnsResolutionData.timestamp = now;
-        socket.send(buf, 0, buf.length, args.port, dnsResolutionData.resolvedAddress, callback);
+        try {
+          socket.send(buf, 0, buf.length, args.port, dnsResolutionData.resolvedAddress, callback);
+        } catch (socketError) {
+          callback(socketError);
+        }
       });
     } else {
       socket.send(buf, 0, buf.length, args.port, dnsResolutionData.resolvedAddress, callback);
--- a/lib/transport.js
+++ b/lib/transport.js
@@ -70,7 +70,11 @@ const createUdpTransport = args => {
         }
       });
     } else {
-      socket.send(buf, 0, buf.length, args.port, dnsResolutionData.resolvedAddress, callback);
+      try {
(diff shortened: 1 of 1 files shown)


lib/transport.js
```diff
@@ -63,7 +63,11 @@ const createUdpTransport = args => {
...
         dnsResolutionData.timestamp = now;
-        socket.send(buf, 0, buf.length, args.port, dnsResolutionData.resolvedAddress, callback);
+        try {
+          socket.send(buf, 0, buf.length, args.port, dnsResolutionData.resolvedAddress, callback);
+        } catch (socketError) {
+          callback(socketError);
+        }
       });
     } else {
       socket.send(buf, 0, buf.length, args.port, dnsResolutionData.resolvedAddress, callback);
```
bdeitte, line 73: Could you change other cases of send so all codpaths now all works the same?

```suggestion
        try {
          socket.send(buf, 0, buf.length, args.port, dnsResolutionData.resolvedAddress, callback);
        } catch (socketError) {
          callback(socketError);
        }
```

bdeitte: There's actually one more below this that I wanted to suggest as well, but GitHub won't let me add in a suggestion for it.

bdeitte:
Nice find!  I'd like to see the other two send() calls fixed for this at the same time, unless I'm missing something there.  Thanks.

lib/transport.js
matteosb: This branch is ok because the exception thrown here is handled at the call site: https://github.com/brightcove/hot-shots/blob/master/lib/statsd.js#L380-L385. The other branch is problematic because the sync exception is thrown in another callback. That said, it looks like wrapping this will result in the same exact behavior, so I'm happy wrap the other socket.sends if you prefer that for consistency.

bdeitte: Thanks, I missed that here.  If you could wrote the other socket I'd appreciate it.

matteosb:
@bdeitte I updated this to just wrap all the socket.send calls for consistency

bdeitte:
Thanks, merging shortly and making a release

PR #272, opened 2024-09-14
Upgrade mocha and package lock for security warning

bdeitte:
Update the non-dev dependencies to fix security warning.  Also a few misc README updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ includes all changes in the latest node-statsd and many additional changes, incl
 * asyncDistTimer
 * much more, including many bug fixes
 
+You can read about all changes in [the changelog](CHANGES.md).
+
 hot-shots supports Node 10.x and higher.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -16,6 +16,8 @@ includes all changes in the latest node-statsd and many additional changes, incl
 * asyncDistTimer
 * much more, including many bug fixes
 
+You can read about all changes in [the changelog](CHANGES.md).
+
```
bdeitte, line 20: One of the most helpful files in the project as it's kept well up to date, so moving it earlier.  And removing two sections that just feel dated below.

package-lock.json
```diff
@@ -1,222 +1,279 @@
 {
   "name": "hot-shots",
   "version": "10.1.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
```
bdeitte, line 4: And now obvious why the tests are failing, still supporting very old Node versions here.  I'll fix up this PR later for this.

Issue #274, opened 2024-10-03
v9.1.0 breaks hierarchical browsing when tags exist


tunesmith:
In v8.3.0, if you incremented a counter of a name, then in graphite-web or grafana, you were still able to browse it hierarchically.

In our case, we also have datadog environment variables. In v9.1.0, those DD_* env variables are automatically added as tags, which is fine, but then it also means that the counter name is added as a tag:name instead, which means the counter name cannot be browsed hierarchically anymore.

I'm fine with having tags, but in the cases where I don't care about the tags while browsing, I'd prefer to be able explore that tag hierarchically in the frontends. Am I missing something basic about how this is intended to work? Can the ability to browse these metric names be hierarchically be restored, even if they were sent with tags?

Issue #203, opened 2021-04-21
Flag to disable DD_ENTITY_ID globalTag behavior


laupow:
`hot-shots` needs a toggle to disable this default DD_ENTITY_ID behavior:

>  The Datadog dd.internal.entity_id tag is appended to globalTags from the DD_ENTITY_ID environment variable if the latter is set

One of our metrics databases was recently overloaded with excessive cardinality because the library starting tagging metrics in a non-Datadog metric pipeline. 

Datadog was enabled on the the application. However, most application metrics went to an Influx database (not Datadog) but still had the `DD_ENTITY_ID` tag. Combined with all our other tags, the database exploded with cardinality. 

We did not expect `hot-shots` to be automatically add this metrics tag.

bdeitte:
Hi @laupow sorry to hear that and thanks for entering the issue.  Note that all hot-shots PR are community-driven at this point, so if you'd like to see this in there, consider creating a PR for it.  Thanks.

PR #275, opened 2024-10-04
Add option 'includeDataDogTags'

thiago-negri:
Adds option `includeDataDogTags` that allows creating a client that ignores the `DD_*` env vars.

Solves:
- https://github.com/brightcove/hot-shots/issues/203
- https://github.com/brightcove/hot-shots/issues/274


tunesmith (approved):

tunesmith:
@bdeitte This seems like a good idea. Some environments have datadog, and also separate statsd integrations that don't want their hierarchical keys squashed as tag:names.

bdeitte:
Makes sense to me- thanks for the addition.  I'll add a minor release with this soon.

PR #276, opened 2024-10-18
Add 'includeDataDogTags' property to 'ClientOptions' type

thiago-negri:
I didn't notice the `ClientOptions` type on #275, so property `includeDataDogTags` is not known for TypeScript projects.

This PR is the attempt to fix it, sorry.


bdeitte:
Ah and I forgot as well, thanks

PR #292, opened 2025-11-29
More override tags and buffer fixups

bdeitte:
This makes one more reverts and fixes related tests

### DIFF
--- a/test/timer.js
+++ b/test/timer.js
@@ -102,6 +102,7 @@ describe('#timer', () => {
       const time = stat.split(/:|\|/)[1];
 
       assert.strictEqual(name, 'name-thingy');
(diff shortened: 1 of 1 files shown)


test/helpers/helpers.js
```diff
@@ -52,7 +52,11 @@ function closeAll(server, statsd, allowErrors, done) {
     statsd.close(() => {
       try {
         if (statsd.hasOwnProperty('protocol') && statsd.protocol === UDS) {
-          server.close(() => { }); // eslint-disable-line no-empty-function
+          server.close((err) => {
+            if (err) {
+              console.log('Error on close:', err);
+            }
+          });
```
bdeitte: Everything here and below is test fixups

PR #321, opened 2026-07-05
Opt-in client-side aggregation and some Datadog client parity changes

bdeitte:
The big change here is a opt-in client-side aggregation of counts, gauges and sets via the `aggregation` option, for better parity with official DogStatsD clients but also for use with any StatsD, DogStatsD or Telegraf client. This includes an aggregation `maxContexts` option (default 5000) bounding the number of live aggregation contexts; new contexts beyond the cap are sent directly with a one-time warning.

A change to help with the aggregation is a a public `flush()` method to send buffered metrics (and any pending aggregated metrics) immediately, useful for serverless and other short-lived environments

The above started when looking at official Datadog client parity, but the aggregation changes widened.  Two changes though for that:
- Support DD_TAGS / DATADOG_TAGS env vars for global tags
- Support DD_DOGSTATSD_URL and DD_DOGSTATSD_SOCKET env vars for transport configuration

And since it showed up while working on these changes, fix dev-dependency security advisories (@babel/core, js-yaml)


bdeitte:
I need to do some more reviewing of this, which likely won't happen until Friday or later, and then this can get merged as a major version update given the number of changes.

Issue #301, opened 2026-02-08
TCP graceful reconnection never triggers due to string vs numeric error code comparison


nayyara-airlangga:
### Overview
TCP socket reconnection via `tcpGracefulErrorHandling` does not work because `protocolErrorHandler()` attempts to compare string error codes (`'EPIPE'`) against an array of numeric errno constants, causing the check to always fail and preventing socket recreation.

### Environment
- **hot-shots version:** 11.2.0
- **Node.js version:** v20.19.2
- **OS:** macOS Sequoia 15.6.1
- **StatsD server version**: prom/statsd-exporter:v0.28.0

### Breakdown

In `lib/statsd.js`, the `handleCallback` function receives socket errors and passes them to `protocolErrorHandler`:
```javascript
const handleCallback = (err, bytes) => {
  // ...
  if (errFormatted) {
    // ...
    if ((this.protocol === PROTOCOL.TCP || this.protocol === PROTOCOL.UDS) && (callback || this.errorHandler)) {
      protocolErrorHandler(this, this.protocol, err);  // err.code is a string like 'EPIPE'
    }
  }
};
```

Then in `protocolErrorHandler`:
```javascript
function protocolErrorHandler(client, protocol, err) {
  // ...
  if (protocol === PROTOCOL.TCP && 
      (!TCP_ERROR_CODES.includes(-err.code) ||  // Comparing -'EPIPE' (NaN) against numeric array
       Date.now() - client.socket.createdAt < client.tcpGracefulRestartRateLimit)) {
    return;  // Always returns here, never recreates socket
  }
  // ... socket recreation code never reached
}
```

To sum up the issue:
- `TCP_ERROR_CODES` (from `constants.tcpErrors()`) contains **numeric** errno values: `[os.constants.errno.EPIPE, ...]` (e.g., `32`)
- Socket errors have `err.code` as a **string**: `'EPIPE'`
- Socket errors do **not** have `err.errno` (it's `undefined`)
- `-err.code` evaluates to `NaN` (negating a string)
- `TCP_ERROR_CODES.includes(NaN)` is always `false`
- `!false` is `true`, so the function **always returns early**

### Sample log

I use Nest.js so here's the log from the service

```
undefined
NaN                                                                                      
[Nest] 97388  - 02/09/2026, 1:28:58 AM   ERROR [MetricService] Error: Error sending hot-shots message: Error: This socket has been ended by the other party                           
[...]

PR #323, opened 2026-07-16
Fix UDS graceful restart for unix-dgram numeric errcodes

KeenanLawrenceStitch:
## Summary

`udsGracefulErrorHandling` never recreates the socket on hot-shots ≥14 when using `unix-dgram`, because `protocolErrorHandler` compares `err.code` against **string** codes only (`'ENOTCONN'`, …), while `unix-dgram` sets `err.code` to a **negative numeric errno** (e.g. `-107` / `-111` on Linux, `-39` / `-54` on Darwin).

This is the UDS follow-on to #301 / #302: the TCP string-code fix was correct for Node TCP sockets, but the same change broke UDS matching.

### Fix

Minimal change in `udsErrors()`: keep the existing string codes and also include the negative numeric errnos that `unix-dgram` emits.

```js
// linux
return ['ENOTCONN', 'ECONNREFUSED', -107, -111];
// darwin
return ['EDESTADDRREQ', 'ECONNRESET', -39, -54];
```

## Test plan

- [x] `npm install`
- [x] `npm test` — 1957 passing
- [x] Updated `test/constants.js` assertions for numeric codes
- [x] CI on this PR

## Related

Follow-up to #301 / #302 (UDS path).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+* Fix UDS graceful restart never firing for unix-dgram errors: `udsErrors()` now includes negative numeric errno codes (`-107`/`-111` on Linux, `-39`/`-54` on Darwin) alongside the existing string codes. unix-dgram sets `err.code` to a negative number (e.g. `-107`), so the string-only check introduced in v14 for #301/#302 could never match and sockets were never replaced after Agent UDS inode churn.
+
 ## 17.0.0 (2026-7-11)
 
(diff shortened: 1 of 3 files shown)


bdeitte:
Thanks- I see as well now this is broken and this looks like a good fix for me.  I'm going to make a few additions here on top of this you'll see here shortly, but this is good to merge.  Thanks again!

KeenanLawrenceStitch:
Awesome, thanks @bdeitte!

PR #329, opened 2026-09-01
Send metrics as utf-8

dylanpulver:
Two places where a non-ascii payload does not survive, one line each.

**Event length header.** `lib/statsFunctions.js:417` (on main) builds `_e{${msgTitle.length},${msgText.length}}`. The [event datagram format](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/) is `_e{<TITLE_UTF8_LENGTH>,<TEXT_UTF8_LENGTH>}:<TITLE>|<TEXT>` — "the length (in bytes) of the UTF-8-encoded" field — while `String#length` counts utf-16 code units. `statsd.event('crème', ...)` announces 5 bytes and then writes 6. `datadogpy` uses `len(title.encode('utf8'))` and `datadog-go` uses `len(event.Title)` on a Go string; both are byte counts.

**TCP transport, now opt-in.** `lib/transport.js:81` (on main) is `socket.write(addEol(buf), 'ascii', ...)`. Following review this is behind a new `tcpEncoding` option that still defaults to `'ascii'`, so the default behaviour is unchanged; `tcpEncoding: 'utf8'` selects the fixed path. Node's `ascii` write encoding is latin1 — only the low byte of each utf-16 code unit survives. On the wire: `montréal` leaves as `montr\xe9al`, `東京` as `q\xac`, `🚀` as `=\x80`. The udp, uds and stream transports all send utf-8, so tcp was the only one.

Measured, same script before and after:

- UDP event, parsed back by slicing the header's byte counts out of the datagram: **5/6 cases fail → 0/6**. The ascii control passes on both sides.
- Raw TCP listener (no `setEncoding`) comparing wire bytes against the utf-8 encoding of the expected line: **4/5 fail → 0/5**.
- The committed tests, run with the project's own `mocha -R spec --timeout 5000 test/*.js`: **1963 passing on `main`, 1974 on the branch, 0 failing on both**. The +11 is the event assertion across the nine transport-matrix rows plus two new TCP tests, one pinning the `ascii` default and one the `utf8` opt-in. No pre-existing test changes status.
- Against a server doing `socket.setEncoding('ascii')`, as `statsd/servers/tcp.js` does, non-ascii is mangled under **both** client encodings (an ascii decode unsets the high bit of every byte), so the only traffic `tcpEncoding: 'utf8'` changes is traffic that already arrived corrupt. Entirely-ascii payloads are byte-identical either way.

Mutants: reverting either line alone leaves the other one's cases failing, so both are load-bearing. Replacing the header with a code-point count (`[...msgTitle].length`) still fails 5/6 — it has to be bytes, not characters.

`test/helpers/helpers.js` set `setEncoding('ascii')` on the tcp test server, so it destroyed non-ascii before any assertion could see it. That is why the new event test needed it switched to utf-8, and it is why the existing suite could not have caught either bug: every event assertion in `test/event.js` is pure ascii, where the two counts coincide.

Not tested: the uds matrix rows — `unix-dgram` aborts on my macOS box, so those were skipped locally; the event change is transport-independent and the uds send path is untouched. No Windows run. I also have not checked how any particular StatsD/DogStatsD server reacts to the previously malformed datagrams — the claim here is only that they did not match the documented format.

Written with Claude Code. To be straight about provenance: I was not debugging a production incident. I was diffing hot-shots' line and event construction against `datadogpy` and `datadog-go` field by field, and these are the two places the three disagree.


lib/transport.js
```diff
@@ -78,7 +78,10 @@ const createTcpTransport = args => {
         }
         return;
       }
-      socket.write(addEol(buf), 'ascii', (err) => {
+      // utf-8, not ascii: node's 'ascii' write encoding keeps only the low byte of
+      // each utf-16 code unit, so a non-ascii metric name, tag or event field would
+      // go out mangled. The udp, uds and stream transports all send utf-8.
+      socket.write(addEol(buf), 'utf8', (err) => {
```
bdeitte: I see this is for TCP, but DogStatsD only supports UDS and UDP- I did an LLM research on this and confirmed this is still the case.  For TCP, this can't just be added in as it would break StatsD (but not Telegraf or OpenTelemetry) so this would need a config value to keep in.

CHANGES.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+* [@dylanpulver](https://github.com/dylanpulver) Fix the DogStatsD event length header to count utf-8 bytes: `_e{...}` carries `TITLE_UTF8_LENGTH` and `TEXT_UTF8_LENGTH`, so a non-ascii title or text made the header disagree with the payload and the event unparseable. Matches `datadogpy` and `datadog-go`.
+* [@dylanpulver](https://github.com/dylanpulver) Add a `tcpEncoding` option. The TCP transport writes with node's `ascii` encoding, which keeps only the low byte of each utf-16 code unit and so corrupts any non-ascii metric name, tag or event field on the wire. That stays the default; set `tcpEncoding: 'utf8'` to send non-ascii intact, as the udp, uds and stream transports already do.
+
```
bdeitte, line 8: I merged in a PR so this needs to be resolved to be included as well here

test/helpers/helpers.js
```diff
@@ -172,15 +172,16 @@ function createServer(serverType, callback) {
   }
   else if (serverType === TCP) {
     server = net.createServer(socket => {
-      socket.setEncoding('ascii');
+      // an ascii decode strips the high bit off every byte, mangling non-ascii.
+      socket.setEncoding('utf8');
```
bdeitte, line 176: I don't want to switch all tests to now use utf8.  Can we find some way where this is just done for specific tests, or it reruns some tests with both options?

lib/statsd.js
```diff
@@ -149,6 +149,8 @@ const Client = function (host, port, prefix, suffix, globalize, cacheDns, mock,
   this.closingFlushInterval = options.closingFlushInterval || 50;
   // Don't set a default type here - let transport.js auto-detect based on host IP version
   this.udpSocketOptions = options.udpSocketOptions || {};
+  // 'ascii' is the historical default and is byte-identical for ascii-only payloads.
+  this.tcpEncoding = options.tcpEncoding || 'ascii';
```
bdeitte, line 153: Suggested fix needed from an LLM review:  ChildClient does not copy tcpEncoding, so a child that ever rebuilds its socket (after a send error, or via trySetNewSocket) silently drops back to ascii for the rest of the process.

lib/transport.js
```diff
@@ -78,7 +78,8 @@ const createTcpTransport = args => {
         }
         return;
       }
-      socket.write(addEol(buf), 'ascii', (err) => {
+      // 'ascii' keeps only the low byte of each utf-16 code unit, so non-ascii goes out mangled.
+      socket.write(addEol(buf), args.tcpEncoding || 'ascii', (err) => {
```
bdeitte, line 82: Huh, so my LLM review is suggesting the option isn't even needed.  Can you see if this would work after updating tests to try both paths?  "The structural suggestion worth considering: sendMessage already has correct utf-8 bytes in a Buffer, and addEol decodes it back to a string only to append a newline, after which socket.write re-encodes it. Writing the Buffer through (Buffer.concat([buf, NEWLINE])) is byte-identical for ascii payloads, makes the header mismatch structurally impossible, and removes the new public option, the README caveat and two allocations per send."

